PR #29, opened 2019-10-15
Refactor for updating method parameter types

AnanyaJha:
### What does this PR do?
This PR fixes the tests related to the changes made for updating the method parameter types

@W-6459892@


package-lock.json
```diff
@@ -0,0 +1,6932 @@
+{
```
lcampos: This project should only be built using `yarn`. Let's delete this file.

src/utils/SfdxCommandBase.ts
```diff
@@ -0,0 +1,48 @@
...
+    const adapter = new ForceGeneratorAdapter();
+    const env = yeoman.createEnv(undefined, undefined, adapter);
+    env.registerStub(generatorname, 'generator');
+
+    const result = await env.run('generator', this.flags);
+    const targetDir = path.resolve(this.flags.outputdir);
+
+    // tslint:disable-next-line:no-unused-expression
+    if (this.flags.json) {
+      return CreateUtil.buildJson(adapter, targetDir);
```
lcampos: Should `buildJson` be part of `SfdxCommandBase` instead of being in the utils class ? I don't think we use it outside of the generator flow.

tsconfig.json
```diff
@@ -1,5 +1,4 @@
 {
-  "extends": "./node_modules/@salesforce/dev-config/tsconfig",
```
lcampos: Why are we removing the `extends` ?

src/utils/SfdxCommandBase.ts
AnanyaJha: It seemed like more of a utility method so I think that's why it was kept in the utils class, but it isn't used outside of the generator flow so I can move it to `SfdxCommandBase`

tsconfig.json
AnanyaJha: I think that might be an old change that crept into this pr, I'll undo it

src/utils/SfdxCommandBase.ts
brpowell: Makes sense to me

src/utils/SfdxCommandBase.ts
```diff
@@ -0,0 +1,48 @@
...
+// tslint:disable-next-line:no-var-requires
+const generator = require('yeoman-generator');
+import { ForceGeneratorAdapter } from './adapter';
+import * as path from 'path';
+import { CreateUtil } from './createUtil';
+// tslint:disable-next-line:no-var-requires
+const yeoman = require('yeoman-environment');
+const messages = Messages.loadMessages('salesforcedx-templates', 'messages');
+
+export class SfdxCommandBase extends SfdxCommand {
```
brpowell: I know this is a nit, but can we call this something more expressive like `TemplateCommand`?

src/utils/SfdxCommandBase.ts
```diff
@@ -0,0 +1,48 @@
...
+  // tslint:disable-next-line:no-any
+  run(): Promise<any> {
+    throw new Error('Method not implemented.');
+  }
+
+  public async runGenerator(generatorname: typeof generator) {
```
brpowell: Another nit but can this also just be `generator`? we're not passing a name which sounds like it would be a string

src/utils/SfdxCommandBase.ts
```diff
@@ -0,0 +1,48 @@
...
+import { ForceGeneratorAdapter } from './adapter';
+import * as path from 'path';
+import { CreateUtil } from './createUtil';
+// tslint:disable-next-line:no-var-requires
+const yeoman = require('yeoman-environment');
+const messages = Messages.loadMessages('salesforcedx-templates', 'messages');
+
+export class SfdxCommandBase extends SfdxCommand {
+  // tslint:disable-next-line:no-any
+  run(): Promise<any> {
```
brpowell: Instead of having this complain at runtime, can we make `run` an abstract method, and the class an abstract one? That way linting/compiling will tell the user to implement it.

test/utils/createUtil.test.ts
```diff
@@ -8,9 +8,9 @@ import { Messages } from '@salesforce/core';
 import { assert, expect } from 'chai';
 import * as fs from 'fs';
 import { resolve } from 'path';
-import { createSandbox, SinonStub, stub } from 'sinon';
-import ApexClassGenerator from '../../src/generators/apexClassGenerator';
+import { SinonStub, stub } from 'sinon';
 import { CreateUtil, ForceGeneratorAdapter, Log } from '../../src/utils';
+import { test } from '@salesforce/command/lib/test';
```
lcampos: This test still references the code that was moved to the template. Lines 94 and 117.

PR #32, opened 2019-10-21
Add Lightning Test Create command

AnanyaJha:
### What does this PR do?
This PR adds the `sfdx force:lightning:test:create` command

@W-6783783@

### DIFF
--- a/test/commands/lightning/test/create.test.ts
+++ b/test/commands/lightning/test/create.test.ts
@@ -26,7 +26,7 @@ describe('Lightning test creation tests:', () => {
         'DefaultLightningTest'
       ])
       .it(
-        'should create lightning test foo using DefaultLightningTest template and aura output directory',
+        'should create lightning test foo using DefaultLightningTest template',
         ctx => {
           assert.file(path.join('foo.resource'));
           assert.file(path.join('foo.resource-meta.xml'));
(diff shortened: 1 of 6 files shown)


AnanyaJha:
Tests still need to be added

test/commands/lightning/test/create.test.ts
```diff
@@ -0,0 +1,142 @@
...
+ */
+import { expect, test } from '@salesforce/command/lib/test';
+import { Messages } from '@salesforce/core';
+import * as path from 'path';
+import * as assert from 'yeoman-assert';
+
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('salesforcedx-templates', 'messages');
+
+describe('Lightning test creation tests:', () => {
```
sfsholden: I think the original description 'Lightning test creation suite' reads better here

test/commands/lightning/test/create.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      .stdout()
+      .command([
+        'force:lightning:test:create',
+        '-n',
+        'foo',
+        '--template',
+        'DefaultLightningTest'
+      ])
+      .it(
+        'should create lightning test foo using DefaultLightningTest template and aura output directory',
```
sfsholden: This test isn't generating the aura output directory like it's described here

test/commands/lightning/test/create.test.ts
lcampos: We should have a test that confirm the `--internal` flag scenario.

AnanyaJha: I agree the double 'test' sounds a little odd, but I left it as is to follow the naming convention used across the other tests in this repo

test/commands/lightning/test/create.test.ts
AnanyaJha: Ah good catch

PR #36, opened 2019-10-23
Loginurl attribute fix for project create

lcampos:
### What does this PR do?
Adds the missing login url attribute to project create. I'm still working on the tests but want to get an early review on the fix.

### DIFF
--- a/src/commands/force/project/create.ts
+++ b/src/commands/force/project/create.ts
@@ -55,7 +55,7 @@ export default class Project extends TemplateCommand {
     }),
     loginurl: flags.string({
       char: 'l',
-      description: 'The login URL for the Salesforce instance being used.', // messages.getMessage('loginurl'),
+      description: messages.getMessage('loginurl'),
       default: 'https://login.salesforce.com',
       hidden: true
     })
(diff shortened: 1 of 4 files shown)


src/commands/force/project/create.ts
```diff
@@ -52,6 +52,12 @@ export default class Project extends TemplateCommand {
     manifest: flags.boolean({
       char: 'x',
       description: messages.getMessage('manifest')
+    }),
+    loginurl: flags.string({
+      char: 'l',
```
lcampos: I'm actually not sure about the character value, anyone knows if this is correct ?

AnanyaJha: Tested it with l and it works!

lcampos: I've confirmed that is the short value for the attribute in the old repo.

src/commands/force/project/create.ts
```diff
@@ -62,7 +68,9 @@ export default class Project extends TemplateCommand {
     this.flags.ns = this.flags.namespace;
 
     // TODO: update the latest apiversion
-    this.flags.sourceApiVersion = '46.0';
+    this.flags.sourceApiVersion = '47.0';
```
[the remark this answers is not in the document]
lcampos: Agreed, I'll open a bug for that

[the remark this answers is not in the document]
lcampos: Done, work item is W-6825719

PR #41, opened 2019-10-31
Make commands compatible with docgen tool

brpowell:
### What does this PR do?

Make commands compatible with docgen tool


src/utils/messageUtil.ts
```diff
@@ -0,0 +1,68 @@
...
+    descriptionKey: string,
+    isLightning: boolean,
+    tokens?: Array<string | number | boolean>
+  ): string {
+    return (
+      this.messages.getMessage(descriptionKey, tokens) +
+      '\n' +
+      this.getHelpHead(isLightning)
+    );
+  }
```
brpowell: The `description` property on commands expected by oclif is equal to: **(old `description`) + (old `help`) - `examples`**

src/utils/messageUtil.ts
```diff
@@ -0,0 +1,68 @@
...
+  public static buildHelpText(
+    examples: string[],
+    isLightning: boolean
+  ): string {
+    return (
+      this.getHelpHead(isLightning) +
+      this.messages.getMessage('help_examples') +
+      examples.reduce((acc, current) => acc + `   ${current}\n`, '')
+    );
+  }
```
brpowell: The old `help` property is equal to: **(oclif `description`) - (old `description`) + `examples`**

src/utils/templateCommand.ts
```diff
@@ -37,26 +38,27 @@ export abstract class TemplateCommand extends SfdxCommand {
...
   public async runGenerator(generator: typeof yeomanGenerator) {
-    // tslint:disable-next-line:no-unused-expression
-    if (!this.flags.apiversion) {
+    const { apiversion, outputdir } = this.flags;
+    if (!apiversion) {
       this.flags.apiversion = TemplateCommand.getDefaultApiVersion();
     }
+    if (outputdir === MessageUtil.get('CurrentWorkingDir')) {
+      this.flags.outputdir = process.cwd();
+    }
```
brpowell: `The current working directory` now shows as the default value, but we need to intercept the flag during generation to actually make it the cwd. We don't want the docs to generate the doc writer's path.

src/commands/force/apex/trigger/create.ts
```diff
@@ -5,38 +5,36 @@
...
     '$ sfdx force:apex:trigger:create -n MyTrigger',
     "$ sfdx force:apex:trigger:create -n MyTrigger -s Account -e 'before insert, after insert'",
     '$ sfdx force:apex:trigger:create -n MyTrigger -d triggers'
   ];
-
-  public static description = messages.getMessage(
-    'ApexTriggerCommandDescription'
-  );
+  public static help = MessageUtil.buildHelpText(ApexTrigger.examples, false);
+  public static longDescription = MessageUtil.get('ApexTriggerLongDescription');
```
brpowell: Properties must be static to be leveraged by the tool

src/utils/messageUtil.ts
```diff
@@ -0,0 +1,68 @@
+import { Messages } from '@salesforce/core';
+
+Messages.importMessagesDirectory(__dirname);
+
+export class MessageUtil {
```
lcampos: Let's add a TODO here that explains this will break the doc plugin once that plugin get updated to oclif.

brpowell:
I have diffed the doc output for each command with the versions generated by FLS. Everything is the same now except for some changes that are likely desirable, like a populated help section for the create commands now.

src/utils/templateCommand.ts
```diff
@@ -37,26 +38,27 @@ export abstract class TemplateCommand extends SfdxCommand {
...
+  public abstract run(): Promise<AnyJson>;
 
   public async runGenerator(generator: typeof yeomanGenerator) {
-    // tslint:disable-next-line:no-unused-expression
-    if (!this.flags.apiversion) {
+    const { apiversion, outputdir } = this.flags;
+    if (!apiversion) {
       this.flags.apiversion = TemplateCommand.getDefaultApiVersion();
     }
+    if (outputdir === MessageUtil.get('OutputDirDefaultDescription')) {
```
lcampos: This misleads what the flag's default value should be in the place that matters the most (the generators). Having this will make it harder to debug & expand the plugin. I suggest we revert the changes to the `outputdir` flag.

PR #42, opened 2019-11-01
added package.json to project

ntotten:
### What does this PR do?
Adds package.json to default project

@W-7236780@

### DIFF
--- a/src/templates/project/.prettierignore
+++ b/src/templates/project/.prettierignore
@@ -2,5 +2,8 @@
 # More information: https://prettier.io/docs/en/ignore.html
 #
 
-.sfdx
 .localdevserver
+.sfdx
+.vscode
+
+coverage/
--- a/src/templates/project/.prettierignore
+++ b/src/templates/project/.prettierignore
@@ -4,6 +4,5 @@
 
 .localdevserver
 .sfdx
-.vscode
 
 coverage/
--- a/src/templates/project/package.json
+++ b/src/templates/project/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "salesforce-app",
+  "private": true,
+  "version": "1.0.0",
+  "description": "Salesforce App",
+  "scripts": {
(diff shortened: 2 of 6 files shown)


src/templates/project/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "salesforce-app",
+  "private": true,
+  "version": "1.0.0",
+  "description": "Salesforce App",
+  "scripts": {
+    "prettier": "prettier --write '**/*.{cls,cmp,component,css,html,js,json,md,page,trigger,yaml,yml}'",
+    "lint": "npm run lint:lwc && npm run lint:aura",
+    "lint:lwc": "eslint **/lwc/**",
+    "lint:aura": "sfdx force:lightning:lint **/aura/** --exit"
```
nrkruk: pretty sure this doesn't work

KrisGraySFDC: Feels like we don't even care about this lint:aura line, we should probably just take it out.

src/templates/project/.prettierignore
```diff
@@ -2,5 +2,8 @@
 # More information: https://prettier.io/docs/en/ignore.html
 #
 
+.localdevserver
 .sfdx
-.localdevserver
\ No newline at end of file
+.vscode
```
smithgp: Any particular reason for not running prettier on the .json files in .vscode/?

KrisGraySFDC: What motivation would you have for that?

smithgp: @KrisGraySFDC Consistent format of those files, particularly settings.json and launch.json, which the users of the generated project might edit manually (since some settings don't have visual editing in VSCode, and launch.json is manually edited).
I have prettier run over my .vscode/ files in all my projects, including our sample analytics template project (https://github.com/forcedotcom/sfdx-analytics)

bpbuch: I assumed these settings were part of the gitignore.  Since they are not, I can see the benefit in consistent formatting. I will remove this line.

PR #43, opened 2019-11-01
Added soql and apex files to standard project

ntotten:
### What does this PR do?
Added soql and apex files to standard project

New project structure:

```
force-app
data
scripts
    apex
       hello.apex
    soql
       accounts.soql
````

### CODE REVIEW

File: src/generators/projectGenerator.ts [hunk 1]
```diff
@@ -123,6 +125,18 @@ export default class ProjectGenerator extends generator {
           )
         )
       );
+      this.fs.copyTpl(
```
lcampos: What if we have both of this under a `scripts` folder so the project structure looks like this

```
- my-app
  - .vscode
  - config
  - force-app
  - manifest
  - scripts
     accounts.soql
     hello.apex
```

src/templates/project/hello.apex
```diff
@@ -0,0 +1,10 @@
+// Use .apex files to store anonymous Apex.
+// You can execute anonymous Apex in VS Code by selecting the
+//     apex text and running the command:
+//     SFDX: Execute Anonymous Apex with Currently Selected Text
+// You can also execute the entire file by running the command:
+//     SFDX: Execute Anonymous Apex with Editor Contents
+
+string tempvar = 'Enter_your_name_here';
+System.debug('Hello World!');
+System.debug('My name is ' + tempvar);
```
praksb: does the language server features work for anonymous apex now?

ntotten: No, but you can still execute anon apex.

src/generators/projectGenerator.ts
ntotten: Scripts feels dangerous because people use it for automation stuff. It could be some other single folder, but not sure what it would be called?

ntotten: https://twitter.com/ntotten/status/1191465122096730114

ntotten: It seems scripts is the winner. :)

src/templates/project/hello.apex
praksb: yes. i think it should be fairly quick (couple of days of work) to enable the language server features for anonymous apex. we just need to pass in a flag to the compiler for parsing. I'll take a stab at it.

src/generators/projectGenerator.ts
```diff
@@ -123,6 +125,18 @@ export default class ProjectGenerator extends generator {
...
+        this.templatePath(path.join(template, soqlQueryFile)),
+        this.destinationPath(
+          path.join(outputdir, projectname, 'scripts/soql', soqlQueryFile)
```
lcampos: This will break on Windows, it needs to be changed to `path.join(outputdir, projectname, 'scripts', 'soql', soqlQueryFile)`

src/generators/projectGenerator.ts
```diff
@@ -123,6 +125,18 @@ export default class ProjectGenerator extends generator {
...
+        )
+      );
+      this.fs.copyTpl(
+        this.templatePath(path.join(template, anonApexFile)),
+        this.destinationPath(
+          path.join(outputdir, projectname, 'scripts/apex', anonApexFile)
```
lcampos: Same as in line 131

test/commands/project/create.test.ts
```diff
@@ -45,6 +45,8 @@ describe('Project creation tests:', () => {
       .command(['force:project:create', '--projectname', 'foo'])
       .it('should create project with default values and foo name', ctx => {
         assert.file([path.join('foo', 'config', 'project-scratch-def.json')]);
+        assert.file([path.join('foo', 'scripts/soql', 'account.soql')]);
```
lcampos: Same situation as described in the `projectGenerator` comments

PR #44, opened 2019-11-02
Project name validation

lcampos:
### What does this PR do?
As part of rewriting the template commands we introduced a validation in the name attribute for the project:create command which caused some regressions. This removes the validation in order to fix the regression.
 
@W-6858055@, #45


.gitignore
```diff
@@ -33,6 +33,9 @@ testsoutput/
 testpage/
 folder space/
 foo*
+analytics1/
```
lcampos: Add missing folders used by tests

package.json
```diff
@@ -27,8 +27,8 @@
     "@types/node": "^10",
     "@types/yeoman-generator": "^3.1.3",
     "chai": "^4",
-    "eslint": "6.0.1",
-    "eslint-config-prettier": "6.0.0",
+    "eslint": "6.6.0",
```
lcampos: Update dev dependency

test/commands/project/create.test.ts
```diff
@@ -153,6 +154,70 @@ describe('Project creation tests:', () => {
           }
         }
       );
+
```
lcampos: Add a test that verifies users can create a project with a dash as part of the name.

test/commands/project/create.test.ts
```diff
@@ -361,69 +426,5 @@ describe('Project creation tests:', () => {
       .it('should throw invalid template name error', ctx => {
         expect(ctx.stderr).to.contain(messages.getMessage('InvalidTemplate'));
       });
-    test
```
lcampos: Remove tests specific to project name validations.

src/commands/force/project/create.ts
```diff
@@ -61,7 +61,6 @@ export default class Project extends TemplateCommand {
     })
   };
   public async run(): Promise<AnyJson> {
-    CreateUtil.checkInputs(this.flags.projectname);
```
lcampos: By removing this validations we now rely on the OS being the one that "checks" if the project name is valid or not, which is what we used to do.

.gitignore
brpowell: I think you can just add `testsoutput/` to cover anything generated by the tests so we don't have to add individual folders.

lcampos: I'll check but I think the ones I added were created outside of `testoutput`

PR #46, opened 2019-11-06
Update @salesforce dependencies to the latest versions

lcampos:
### What does this PR do?
Updates the plugin to use the latest @salesforce dependencies to address some reported vulnerabilities.

@W-6874896@


package.json
```diff
@@ -10,8 +10,8 @@
     "@oclif/command": "^1",
     "@oclif/config": "^1",
     "@oclif/errors": "^1",
-    "@salesforce/command": "^1.4.1",
-    "@salesforce/core": "^1.3.2",
+    "@salesforce/command": "2.1.3",
```
tnoonan-salesforce: Can you do ^2 for both?

tnoonan-salesforce: We stick to semver so you should be safe picking up minor features and patches.

yarn.lock
```diff
@@ -180,34 +191,47 @@
...
-    "@oclif/errors" "^1.2.0"
-    "@oclif/parser" "^3.6.0"
-    "@oclif/plugin-help" "^2.1.0"
-    "@oclif/test" "^1.2.0"
-    "@salesforce/core" "^1.2.3"
-    "@salesforce/kit" "^1.1.0"
-    "@salesforce/ts-types" "^1.0.0"
-    chalk "^2.4.0"
-    cli-ux "^4.8.0"
+"@salesforce/command@2.1.3":
```
brpowell: I think the lock file needs to be regenerated with the ^2 changes.

lcampos: Ah right, good catch 👍

PR #47, opened 2019-11-06
Project create to use defaultApiVersion method for setting version

brpowell:
### What does this PR do?

Updates project:create to use the api version method to get the version instead of relying on the hardcoded value 47.0.

W-6825719

### DIFF
--- a/src/commands/force/project/create.ts
+++ b/src/commands/force/project/create.ts
@@ -75,11 +75,6 @@ export default class Project extends TemplateCommand {
     // namespace is a reserved keyword for the generator
     this.flags.ns = this.flags.namespace;
 
-    // TODO: update the latest apiversion
-    this.flags.sourceApiVersion = '47.0';
-
-    this.flags.loginURL = this.flags.loginurl;
-
     return this.runGenerator(ProjectGenerator);
   }
 }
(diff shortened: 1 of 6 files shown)


src/commands/force/project/create.ts
```diff
@@ -75,11 +75,6 @@ export default class Project extends TemplateCommand {
     // namespace is a reserved keyword for the generator
     this.flags.ns = this.flags.namespace;
 
-    // TODO: update the latest apiversion
-    this.flags.sourceApiVersion = '47.0';
-
-    this.flags.loginURL = this.flags.loginurl;
```
brpowell: We can remove the fake flag where we set the api version and just have the TemplateCommand base class take care of setting it like we do for all the other commands. Also updating the parameters in the templates to use `apiversion` and `loginurl` as the names so we don't have to rename them here.

src/templates/project/sfdx-project.json
```diff
@@ -1,11 +1,11 @@
...
\ No newline at end of file
+  "packageDirectories": [
+    {
+      "path": "<%= defaultpackagedir %>",
+      "default": true
+    }
+  ],
+  "namespace": "<%= namespace %>",
+  "sfdcLoginUrl": "<%= loginurl %>",
+  "sourceApiVersion": "<%= apiversion %>"
```
brpowell: Changing the parameter names to the flag names

PR #52, opened 2019-11-12
Ensure LWC class name is the same as the value passed with the name flag

brpowell:
### What does this PR do?

Update LWC generation so that only the filenames are camelCased, while the class name stays consistent with what the user enters for the name flag.

### DIFF
--- a/src/generators/lightningComponentGenerator.ts
+++ b/src/generators/lightningComponentGenerator.ts
@@ -95,27 +95,30 @@ export default class LightningComponentGenerator extends generator {
     }
     // tslint:disable-next-line:no-unused-expression
     if (type === 'lwc') {
+      // lwc requires first letter of filename to be lowercase
+      const fileName = `${componentname
+        .substring(0, 1)
+        .toLowerCase()}${componentname.substring(1)}`;
+
       this.sourceRoot(
         path.join(__dirname, '..', 'templates', 'lightningcomponent', 'lwc')
       );
       this.fs.copyTpl(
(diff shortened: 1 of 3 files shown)


src/generators/lightningComponentGenerator.ts
```diff
@@ -95,27 +95,30 @@ export default class LightningComponentGenerator extends generator {
...
       this.sourceRoot(
         path.join(__dirname, '..', 'templates', 'lightningcomponent', 'lwc')
       );
       this.fs.copyTpl(
         this.templatePath('DefaultLightningLWC.js'),
-        this.destinationPath(
-          path.join(outputdir, componentname, `${componentname}.js`)
-        ),
+        this.destinationPath(path.join(outputdir, fileName, `${fileName}.js`)),
         { componentname }
```
brpowell: We keep the original componentname flag value here when passing it to the template

src/generators/lightningComponentGenerator.ts
```diff
@@ -95,29 +95,32 @@ export default class LightningComponentGenerator extends generator {
...
           )
         );
       if (!internal) {
         this.fs.copyTpl(
           this.templatePath('_js-meta.xml'),
           this.destinationPath(
-            path.join(outputdir, componentname, `${componentname}.js-meta.xml`)
+            path.join(outputdir, fileName, `${fileName}.js-meta.xml`)
           ),
-          { apiVersion: apiversion, componentname }
```
brpowell: this template file didn't use `componentname` at all.

Issue #53, opened 2019-11-13
LWC create command prevents class name being uppercase


lcampos:
Running `sfdx force:lightning:component:create --type lwc --componentname WhiteLimo --outputdir force-app/main/default/lwc` will create a component with a class file that has the name is camelCase.

### Steps To Reproduce:

Run the lightning create command with a component name that starts with a capital letter.

### Expected result

The class name in the js file should be pascal case.

### Actual result

The class name for the component is camelCase.

**SFDX CLI Version**: 7.32
**OS and version**: Windows & Mac OS

Source: https://twitter.com/damianpoole/status/1194333018120474624?s=20

lcampos:
Closing since this was fixed and released as part of the Salesforce CLI version 7.33.2

PR #54, opened 2019-11-13
LWC js class name should always be pascal case

lcampos:
### What does this PR do?
Follow up to PR #52 to address scenarios where `--componentname` value is camelCase and js class name should be pascal case.

#53  , @W-6902712@

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,17 +1,16 @@
 {
-    "files.exclude": {
-      "**/.git": true,
-      "**/.svn": true,
-      "**/.hg": true,
-      "**/CVS": true,
-      "**/.DS_Store": true
-    },
-    "search.exclude": {
-      "**/lib": true,
-      "**/bin": true
-    },
-    "editor.tabSize": 2,
-    "editor.formatOnSave": true,
-    "prettier.singleQuote": true,
-    "rewrap.wrappingColumn": 80
-  }
+  "files.exclude": {
+    "**/.git": true,
+    "**/.svn": true,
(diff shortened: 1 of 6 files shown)


src/commands/force/lightning/component/create.ts
```diff
@@ -83,6 +83,13 @@ export default class LightningComponent extends TemplateCommand {
     CreateUtil.checkInputs(this.flags.componentname);
     CreateUtil.checkInputs(this.flags.template);
 
+    if (this.flags.type === 'lwc') {
+      // lwc's convention is for the class name to be Pascal Case
+      this.flags.componentname = `${this.flags.componentname
```
lcampos: Based on the docs (https://developer.salesforce.com/docs/component-library/documentation/lwc/lwc.create_components_javascript) and the old code base the js class name should always be pascal case.

.vscode/settings.json
```diff
@@ -1,17 +1,16 @@
...
-      "**/CVS": true,
-      "**/.DS_Store": true
-    },
-    "search.exclude": {
-      "**/lib": true,
-      "**/bin": true
-    },
-    "editor.tabSize": 2,
-    "editor.formatOnSave": true,
-    "prettier.singleQuote": true,
```
lcampos: Removed `prettier.singleQuote` since it got deprecated.

src/commands/force/lightning/component/create.ts
brpowell: Can we do this conversion in the generator so that all the casing modifications can be done in the same place?

PR #62, opened 2020-01-10
Duplicate project creation story

jag-sfdc:
### What does this PR do?
Currently "sfdx force:project:create" creates a duplicate empty project structure in the working directory where the command was issued. This PR fixes that issues.

W-7094617

### DIFF
--- a/src/commands/force/project/create.ts
+++ b/src/commands/force/project/create.ts
@@ -41,7 +41,7 @@ export default class Project extends TemplateCommand {
       char: 'd',
       description: MessageUtil.get('OutputDirFlagDescription'),
       longDescription: MessageUtil.get('OutputDirFlagLongDescription'),
-      default: process.cwd()
+      default: '.'
     }),
     namespace: flags.string({
       char: 's',
--- a/src/generators/projectGenerator.ts
+++ b/src/generators/projectGenerator.ts
@@ -185,8 +185,8 @@ function makeEmptyFolders(
   for (const folder of toplevelfolders) {
     if (!fs.existsSync(path.join(oldfolder, folder))) {
       fs.mkdirSync(path.join(oldfolder, folder));
-      oldfolder = path.join(oldfolder, folder);
     }
+    oldfolder = path.join(oldfolder, folder);
   }
   for (const newfolder of metadatafolders) {
     if (!fs.existsSync(path.join(oldfolder, newfolder))) {
--- a/test/commands/project/create.test.ts
+++ b/test/commands/project/create.test.ts
@@ -161,6 +161,27 @@ describe('Project creation tests:', () => {
       .withOrg()
       .withProject()
       .stdout()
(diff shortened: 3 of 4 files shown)


test/commands/project/create.test.ts
```diff
@@ -157,6 +157,27 @@ describe('Project creation tests:', () => {
         }
       );
 
+    test
```
jag-sfdc: Added a Red/Green test to reproduce the issue.

src/generators/projectGenerator.ts
```diff
@@ -185,8 +185,8 @@ function makeEmptyFolders(
   for (const folder of toplevelfolders) {
     if (!fs.existsSync(path.join(oldfolder, folder))) {
       fs.mkdirSync(path.join(oldfolder, folder));
-      oldfolder = path.join(oldfolder, folder);
```
jag-sfdc: The issue is a little different that originally thought. It does not create an entirely duplicate project. But instead creates half the project structure in the intended target directory and rest of the directories in the duplicate project.

src/commands/force/project/create.ts
```diff
@@ -41,7 +41,7 @@ export default class Project extends TemplateCommand {
       char: 'd',
       description: MessageUtil.get('OutputDirFlagDescription'),
       longDescription: MessageUtil.get('OutputDirFlagLongDescription'),
-      default: process.cwd()
```
lcampos: Seems like this branch was created from the `project_creation_outputdir_story` branch and not from `master`.

test/commands/project/create.test.ts
brpowell: I was only able to reproduce this if the output directory already existed. Do you observe that too? If it doesn't, it seems to work as expected. Should the test verify "test outputdir" is already present, and if it isn't create it before running?

jag-sfdc: I saw some flakiness with this test too. But after deleting the "testsoutput/" from the previous runs, I was able to consistently make it pass with the fix (I also commented out the fix and made sure it fails consistently).

brpowell: I'm wondering if we should clear the testsoutput folder before running the test suite. We could minimize the situation of If I was trying to fix a failure and was getting false negatives because of leftover files from previous runs.

jag-sfdc: Sure, I'm for it. It is a little bit more tricky since a few tests I see create project outside of the "testsoutput" directory. Can we do it as a separate story and not increase the scope of this story.

lcampos: @brpowell @jag-sfdc lets just clear the test generated folders that are specific to the new test introduced here. I'll add a separate work item to address this issue at the repository level.

brpowell: Sounds good 👍

PR #71, opened 2020-01-15
Jj/port duplicate project bug fix to release @W-7094617@

jag-sfdc:
### What does this PR do?
Currently "sfdx force:project:create" creates a duplicate empty project structure in the working directory where the command was issued. This PR fixes that issues.

W-7094617

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ To build the plugin locally, make sure to have yarn installed and run the follow
 
 ```
 Clone the repository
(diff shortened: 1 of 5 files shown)


jag-sfdc:
I don't understand why this pull request included two other commits that are already in master.

brpowell:
I think your local release branch needs to be updated and then base the port branch off of that. And the PR needs to target release not master.

PR #83, opened 2020-02-26
Windows build/test using CircleCI

rcoringrato-sfdc:
### What does this PR do?

- Add build/test job for windows
- Process and upload code coverage

@W-7129488@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,8 +27,7 @@ jobs:
       - run:
           name: Submitting code coverage to codecov
           command: |
-            ./node_modules/.bin/nyc report --reporter text-lcov > coverage.lcov
-            bash <(curl -s https://codecov.io/bash) -f "coverage.lcov" -t $CODECOV_KEY
+            codecov --disable=gcov
   node-8:
     <<: *test
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,3 @@
-#
-# CircleCI configuration for salesforcedx-templates
-#
 ---
 version: 2.1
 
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -47,6 +46,41 @@ jobs:
...
 
+  run-win-tests:
+    description: 'Run tests on Windows'
+    executor: win/default
+    steps:
+      - run: git config --global core.autocrlf false
+      - checkout
+      - restore_cache:
+          keys:
+            - dependency-cache-win1-{{checksum ".circleci/config.yml"}}-{{checksum "yarn.lock"}}
```
lcampos: What's the benefit of including the one for the circleci config ? Isn't the checksum for yarn.lock enough ?

.circleci/config.yml
```diff
@@ -47,6 +46,41 @@ jobs:
...
+            # if cache for exact version of 'package.json' is not present, then load the most recent cache
+            - dependency-cache-win1-
+      - run:
+          name: 'Install dependencies'
+          command: |
+            yarn --version
+            node --version
+            yarn
+            npm install -g codecov
```
lcampos: Can we change this to `yarn global ...` or is there a reason why it needs to be installed using npm ?

package.json
```diff
@@ -69,7 +69,7 @@
   },
   "repository": "forcedotcom/salesforcedx-templates",
   "scripts": {
-    "clean:lib": "node scripts/remove-lib",
+    "clean:lib": "node scripts/remove-lib && shx rm -rf coverage && shx rm -rf .nyc_output",
```
lcampos: Not a thing we need to address in this PR but should we get rid of the `remove-lib` script and replace it with `shx` ? If so I can create a work item for it.

.circleci/config.yml
rcoringrato-sfdc: @lcampos A benefit is that you start with a clean cache when you make circleci config changes.

.circleci/config.yml
rcoringrato-sfdc: @lcampos No reason other than habit. I'll change it.

package.json
rcoringrato-sfdc: @lcampos Straight forward change, I'll include in this PR.

.circleci/config.yml
lcampos: Got it 👍

PR #87, opened 2020-02-26
force:analytics updates

smithgp:
### What does this PR do?
- remove undocumented `storedQueries` field from sample template-info.json
- format sample template files to match default prettier format
- add oclif topic description for force:analytics and force:analytics:template

- W-7258830 (storedQueries)
- W-7222866 (topic description)

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,17 @@
     "bin": "sfdx",
     "devPlugins": [
       "@oclif/plugin-help"
-    ]
+    ],
+    "topics": {
+      "force:analytics": {
+        "description": "work with analytics assets",
+        "subtopics": {
+          "template": {
+            "description": "work with analytics templates"
+          }
+        }
+      }
+    }
   },
   "repository": "forcedotcom/salesforcedx-templates",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -65,7 +65,17 @@
     "bin": "sfdx",
     "devPlugins": [
       "@oclif/plugin-help"
-    ]
+    ],
+    "topics": {
```
lcampos: Outside of this change to package.json the rest seems to be formatting changes to the template. Is this correct ?

smithgp: @lcampos Actually, the first commit does remove the `storedQueries: []` field from the template-info.json. The 2nd commit is all formatting of the template files, plus the updates to the unit test to account for the difference in spacing.

smithgp: You use the Hide whitespace changes checkbox in the gear icon on Files changed tab to see the real changes better.

lcampos: Cool, I missed that `storedQueries` change since I was going through this looking at all the changes at once.

PR #103, opened 2020-03-11
Add command reference and deprecation tests

amphro:
### What does this PR do?
Adds tests for command reference generation and deprecation to catch issues at dev and CI time v.s. CLI publish time.

@W-7319588@

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,11 +7,9 @@
     {
       "type": "node",
       "request": "attach",
-      "name": "Attach to Remote",
-      "address": "127.0.0.1",
+      "name": "Attach",
       "port": 9229,
-      "localRoot": "${workspaceFolder}",
-      "remoteRoot": "${workspaceFolder}"
+      "skipFiles": ["<node_internals>/**"]
     },
     {
       "name": "Run All Tests",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -64,28 +66,56 @@
...
-      "@oclif/plugin-help"
+      "@oclif/plugin-help",
+      "@oclif/plugin-command-snapshot",
+      "@salesforce/plugin-command-reference"
     ],
     "topics": {
-      "force:analytics": {
-        "description": "work with analytics assets",
+      "force": {
+        "external": true,
```
smithgp: Nice, I didn't know about the `external: true` thing.

package.json
```diff
@@ -5,26 +5,28 @@
...
     "@oclif/dev-cli": "^1",
+    "@oclif/plugin-command-snapshot": "^1.0.0",
     "@oclif/plugin-help": "^2",
     "@oclif/test": "^1",
     "@salesforce/dev-config": "1.4.1",
+    "@salesforce/plugin-command-reference": "^1.0.7",
     "@types/chai": "^4",
     "@types/mocha": "^5",
-    "@types/node": "^10",
+    "@types/node": "^13.9.0",
```
smithgp: The .nvmrc for this has node version at 10.15.3, and the engine says >= 8.10.
Should those be updated, too?

.vscode/launch.json
```diff
@@ -7,11 +7,9 @@
     {
       "type": "node",
       "request": "attach",
-      "name": "Attach to Remote",
```
lcampos: Do these changes still allow us to attach the Node debugger ?

PR #118, opened 2020-04-22
Update imports and issues related with types

lcampos:
### What does this PR do?
Updates how we import certain dependencies. Fixes issues surfaced by the import bringing type information.

@W-7481028@

### DIFF
--- a/src/generators/apexClassGenerator.ts
+++ b/src/generators/apexClassGenerator.ts
@@ -5,14 +5,14 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import * as path from 'path';
+import * as Generator from 'yeoman-generator';
 import { OptionsMap } from '../utils/types';
-// tslint:disable-next-line:no-var-requires
-const generator = require('yeoman-generator');
 
-export default class ApexClassGenerator extends generator {
+export default class ApexClassGenerator extends Generator {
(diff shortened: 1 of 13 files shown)


src/generators/apexClassGenerator.ts
```diff
@@ -5,14 +5,14 @@
...
 import { OptionsMap } from '../utils/types';
-// tslint:disable-next-line:no-var-requires
-const generator = require('yeoman-generator');
 
-export default class ApexClassGenerator extends generator {
+export default class ApexClassGenerator extends Generator {
   constructor(args: string | string[], options: OptionsMap) {
     super(args, options);
     this.sourceRoot(path.join(__dirname, '..', 'templates', 'apexclass'));
+    // @ts-ignore
```
lcampos, line 15: Looks like `conflicter` is not part of the `Generator` type. @AnanyaJha by any chance do you remember if this was part of the changes needed for the logger code ?

AnanyaJha: Hmm from what I remember Amanda had included this so we wouldn't confirm with the user before overwriting their files, and we changed it to false later on because we decided that we wanted that functionality. I'm not totally sure if this is a new change, but it looks like Conflicter is now its own class https://yeoman.github.io/generator/Conflicter.html

lcampos: Ok, I'll log a separate work item to look into it.

Issue #110, opened 2020-03-14
Add developer productivity settings to `project-scratch-def.json`


muenzpraeger:
**Is your feature request related to a problem? Please describe.**
The scratch org definition file for new projects should per default:
- allow to set the password via API (needed for `sfdx force:user:password:generate`)
- disable UX caching in Lightning
- force LEX enablement

**Describe the solution you'd like**
Add [these settings](https://github.com/trailheadapps/lwc-recipes/blob/master/config/project-scratch-def.json#L5-L16) to the `project-scratch-def.json` template file:

PR #126, opened 2020-05-12
Updates for analytics project and sample template

smithgp:
- Add the dev productivity settings (from #110) to the analytics project scratch-def.json
- Hardcode the assetVersion in the analytics to 48.0 to match the format of the dashboard in the template
- Include the `_tp` name suffix on the sample analytics template dashboard, to support it being used in multiple embedded analytics app in the server

### DIFF
--- a/src/templates/analytics/waveTemplates/DefaultAnalyticsTemplate/template-info.json
+++ b/src/templates/analytics/waveTemplates/DefaultAnalyticsTemplate/template-info.json
@@ -3,7 +3,7 @@
   "label": "<%= templateName %>",
   "name": "<%= templateName %>",
   "description": "",
-  "assetVersion": <%= sourceApiVersion %>,
+  "assetVersion": 48.0,
   "variableDefinition": "variables.json",
   "uiDefinition": "ui.json",
   "rules": [
--- a/src/templates/analytics/waveTemplates/DefaultAnalyticsTemplate/template-info.json
+++ b/src/templates/analytics/waveTemplates/DefaultAnalyticsTemplate/template-info.json
@@ -26,7 +26,7 @@
   "dashboards": [
     {
       "label": "<%= templateName %>Dashboard",
-      "name": "<%= templateName %>Dashboard",
+      "name": "<%= templateName %>Dashboard_tp",
       "condition": "${Variables.Overrides.createAllDashboards}",
       "file": "dashboards/<%= templateName %>Dashboard.json"
     }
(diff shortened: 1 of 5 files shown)


test/commands/project/create.test.ts
```diff
@@ -362,7 +362,7 @@ describe('Project creation tests:', () => {
           }
           assert.fileContent(
             path.join('analytics1', 'config', 'project-scratch-def.json'),
-            '["DevelopmentWave"]'
+            '"DevelopmentWave"'
```
smithgp, line 365: When the ScratchDef.json got prettier-ized, it added newlines in the `features` array, so this test was mismatching. We really only care that `"DevelopmentWave"` is in there somewhere.

src/templates/analytics/waveTemplates/DefaultAnalyticsTemplate/template-info.json
```diff
@@ -3,7 +3,7 @@
   "label": "<%= templateName %>",
   "name": "<%= templateName %>",
   "description": "",
-  "assetVersion": <%= sourceApiVersion %>,
+  "assetVersion": 48.0,
```
smithgp, line 6: Also, the git pre-commit husky hook that got added to reformat all json files was barfing on this since the raw file before was not valid json (with the <%= ... %> being used to insert a number in there), so it's serendipitous that we needed to change this anyways.

smithgp: @ntotten @lcampos This means that, until this change goes in, no one will be able to commit locally (unless they use --no-verify).

brpowell: It looks like the templates folder was added to the prettierignore file a few hours ago. Can you rebase and see if that solves this problem? Preferable to not have to hardcode the api version.

smithgp: Actually, we do need to hardcode the assetVersion, for reasons unrelated to the git hook reformatting. The assetVersion in an analytics template isn't really the same as the sourceApiVersion -- assetVersion describes the format of the other files in the template, which can be different than the sourceApiVersion.
And, I do want to prettier-up these files (now that they'll be valid json again), so I'll rebase and take the templates folder back out of the .prettierignore.

smithgp: And, that's weird -- turns out I already had Nate's change that updated the .prettierignore, but my git commit still failed on parsing this file (via the pre-commit hook from husky) until I removed the <%= ... =>

PR #140, opened 2020-07-02
Allow api version to be set by config

mdonnalley:
### What does this PR do?
Allows users to specify api versions by either using `SFDX_API_VERSION` env var or setting the `apiVersion` via `sfdx force:config:set`

N/A

### DIFF
--- a/src/utils/templateCommand.ts
+++ b/src/utils/templateCommand.ts
@@ -6,6 +6,7 @@
  */
 
 import { SfdxCommand } from '@salesforce/command';
+import { ConfigAggregator } from '@salesforce/core';
 import { AnyJson } from '@salesforce/ts-types';
 import * as path from 'path';
 // @ts-ignore
@@ -16,6 +17,8 @@ import { MessageUtil } from './messageUtil';
 import { CreateOutput } from './types';
 
 export abstract class TemplateCommand extends SfdxCommand {
+  private static API_VERSION = 'apiVersion';
+
   public static buildJson(
     adapter: ForceGeneratorAdapter,
     targetDir: string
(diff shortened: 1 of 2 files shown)


src/utils/templateCommand.ts
```diff
@@ -30,18 +33,25 @@ export abstract class TemplateCommand extends SfdxCommand {
...
   }
 
   public abstract run(): Promise<AnyJson>;
 
   public async runGenerator(generator: typeof yeomanGenerator) {
     // Can't specify a default value the normal way for apiversion, so set it here
-    if (!this.flags.apiversion) {
-      this.flags.apiversion = TemplateCommand.getDefaultApiVersion();
-    }
+    this.flags.apiversion = await TemplateCommand.getApiVersion();
```
sfsholden, line 54: I think we'd want to keep the conditional check here. I don't think it'd be worth resetting `this.flags.apiversion` repeatedly. The rest looks good to me.

src/utils/templateCommand.ts
```diff
@@ -30,18 +33,25 @@ export abstract class TemplateCommand extends SfdxCommand {
     return output;
   }
 
-  public static getDefaultApiVersion(): string {
-    const versionTrimmed = require('../../package.json').version.trim();
-    return `${versionTrimmed.split('.')[0]}.0`;
+  public static async getApiVersion(): Promise<string> {
+    const aggregator = await ConfigAggregator.create();
```
lcampos, line 37: I think this needs to go inside a try/catch

yarn.lock
```diff
@@ -5526,6 +5763,40 @@ yeoman-generator@^4.7.2:
     mem-fs-editor "^6.0.0"
     yeoman-environment "^2.3.4"
 
+yeoman-generator@^4.8.2:
```
lcampos, line 5766: Did you have to add a new import to the project or why is the lock file getting updated ? I think some of these changes might break the CLI build.

mdonnalley: @lcampos I didn't have to import anything. The yarn.lock is updated every time I run `yarn`... I can easily commit the yarn.lock from master if you prefer. I just assumed that the yarn.lock was automatically updated for a reason

mdonnalley:
@lcampos I went ahead and checked in the yarn.lock from master. Let me know if you want me to make any other changes

PR #142, opened 2020-07-08
add additional settings for js-meta.xml support

a2975667:
### What does this PR do?
This PR add additional default settings to workspace settings for js-meta.xml intellisense support.

This PR is associated with [this PR](https://github.com/forcedotcom/salesforcedx-vscode/pull/2314).
This PR is also associated with work item @W-7737982@

### DIFF
--- a/src/templates/project/settings.json
+++ b/src/templates/project/settings.json
@@ -3,5 +3,14 @@
     "**/node_modules": true,
     "**/bower_components": true,
     "**/.sfdx": true
-  }
+  },
+  "xml.catalogs": [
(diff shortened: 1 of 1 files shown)


ntotten:
I don't think this is a very good solution. Can we see if we can send a PR to the red hat extension to make this something our extension can register directly rather than have to put it in everyone's settings? https://github.com/redhat-developer/vscode-xml

a2975667:
Closing this PR since RedHat XML now supports APIs to setup the extension without using settings.json.

PR #147, opened 2020-07-29
Refactor into monorepo

xyc:
### What does this PR do?
Refactor package into monorepo
- [x] separate into `plugin-templates` and `templates` packages 
- [x] update npm scripts: build, lint, test
- [x] fix tests
- [x] use i18n for `templates` package
- [x] clean up duplicate utils and messages in both packages

@W-7746937@


lerna.json
```diff
@@ -0,0 +1,5 @@
+{
+  "version": "0.0.1",
```
lcampos, line 2: We'll need to change this to 49.x.x so we can keep the library and the plugin at the same version. A similar change would be needed on each module's package.json version field.

packages/templates/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "tslib": "^1",
+    "yeoman-environment": "2.4.0",
+    "yeoman-generator": "4.0.1"
+  },
+  "devDependencies": {
+    "@types/node": "^10",
+    "@types/yeoman-generator": "^3.1.3",
+    "shx": "0.2.2",
+    "tslint": "^6.1.0",
+    "typescript": "^3.8.3"
```
lcampos, line 25: Is this version of typescript compatible with the one we're using in the vscode repo ?

xyc: Good question, I didn't change the original version. Let me test it out.

lerna.json
xyc: Updated all packages to 49.0.0

packages/templates/package.json
xyc: Tested out on VS Code `npm pack`ed package and it looks good

PR #149, opened 2020-08-04
Library API

xyc:
### What does this PR do?
Define library API: https://salesforce.quip.com/FQjvAkt4W98q 
- [x] generic create template API
- [x] update `@types/yeoman-generator` version
- [x] Define new base generator class `SfdxGenerator`, refactoring and define patterns for option validation
- [x] test CLI with new API. branch: [xyc/node-api-test-cli](https://github.com/forcedotcom/salesforcedx-templates/compare/xyc/node-api...xyc/node-api-test-cli?expand=1). Tested `force:apex:class:create` and `force:lightning:test:create` with new API - all tests pass.
- [x] Update existing generators to extend `SfdxGenerator` and migrate validation code
- [x] Unit tests

@W-7746897@

### DIFF
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -289,7 +289,7 @@ OPTIONS
 
   -s, --namespace=namespace                                                         project associated namespace
 
-  -t, --template=standard|empty                                                     [default: standard] template to use
+  -t, --template=standard|empty|analytics                                           [default: standard] template to use
                                                                                     for file creation
 
   -x, --manifest                                                                    generate a manifest (package.xml)
--- a/packages/templates/package.json
+++ b/packages/templates/package.json
@@ -19,7 +19,8 @@
   },
   "devDependencies": {
     "@types/node": "^10",
-    "@types/yeoman-generator": "^3.1.3",
+    "@types/yeoman-environment": "2.10.0",
+    "@types/yeoman-generator": "3.1.3",
     "shx": "0.2.2",
     "tslint": "^6.1.0",
     "typescript": "^3.8.3"
--- a/packages/templates/package.json
+++ b/packages/templates/package.json
@@ -18,18 +18,31 @@
     "yeoman-generator": "4.0.1"
   },
   "devDependencies": {
+    "@types/chai": "^4",
+    "@types/fs-extra": "^8.1.0",
+    "@types/mocha": "^5",
     "@types/node": "^10",
+    "@types/sinon": "^7.3.1",
+    "@types/yeoman-assert": "^3.1.1",
     "@types/yeoman-environment": "2.10.0",
     "@types/yeoman-generator": "4.11.0",
+    "chai": "^4",
+    "fs-extra": "^9.0.1",
+    "mocha": "^6.2.3",
+    "nyc": "15.0.1",
+    "sinon": "^7.3.1",
     "shx": "0.2.2",
+    "ts-node": "^8.6.2",
     "tslint": "^6.1.0",
-    "typescript": "^3.8.3"
+    "typescript": "^3.8.3",
+    "yeoman-assert": "^3.1.1"
   },
   "scripts": {
     "clean:lib": "shx rm -rf lib && shx rm -rf coverage && shx rm -rf .nyc_output",
     "compile": "tsc -b",
     "build": "yarn run clean:lib && yarn build:templates && yarn compile",
     "build:templates": "node scripts/build-templates",
+    "test": "nyc --exclude src/i18n/localization.ts --extension .ts mocha --forbid-only \"test/**/*.test.ts\"",
     "lint": "tslint --project . --config tslint.json --format stylish"
   }
 }
--- a/packages/templates/package.json
+++ b/packages/templates/package.json
@@ -20,7 +20,7 @@
   "devDependencies": {
     "@types/node": "^10",
     "@types/yeoman-environment": "2.10.0",
-    "@types/yeoman-generator": "3.1.3",
+    "@types/yeoman-generator": "4.11.0",
     "shx": "0.2.2",
     "tslint": "^6.1.0",
     "typescript": "^3.8.3"
--- a/packages/templates/src/generators/sfdxGenerator.ts
+++ b/packages/templates/src/generators/sfdxGenerator.ts
@@ -1,0 +1,29 @@
+import * as path from 'path';
+import * as Generator from 'yeoman-generator';
+import { TemplateOptions } from '../utils/types';
+
+export abstract class SfdxGenerator<
+  TOptions extends TemplateOptions
+> extends Generator<TOptions> {
+  options!: TOptions & {
+    apiversion: string;
+    outputdir: string;
+  };
+  constructor(args: string | string[], options: TOptions) {
+    super(args, options);
+    this.options.apiversion =
+      this.options.apiversion ?? this.getDefaultApiVersion();
+    this.options.outputdir = this.options.outputdir ?? process.cwd();
+    this.validateOptions();
+  }
+  public abstract validateOptions(): void;
+
+  /**
+   * Look up package version of @salesforce/templates package to supply a default API version
+   */
+  private getDefaultApiVersion(): string {
+    const packageJsonPath = path.join('..', '..', 'package.json');
+    const versionTrimmed = require(packageJsonPath).version.trim();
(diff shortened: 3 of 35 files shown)


packages/templates/src/generators/sfdxGenerator.ts
```diff
@@ -0,0 +1,29 @@
+import * as path from 'path';
```
lcampos, line 1:
```suggestion
/*
 * Copyright (c) 2020, salesforce.com, inc.
 * All rights reserved.
 * Licensed under the BSD 3-Clause license.
 * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
 */
import * as path from 'path';
```

packages/templates/src/generators/sfdxGenerator.ts
```diff
@@ -0,0 +1,29 @@
...
+    outputdir: string;
+  };
+  constructor(args: string | string[], options: TOptions) {
+    super(args, options);
+    this.options.apiversion =
+      this.options.apiversion ?? this.getDefaultApiVersion();
+    this.options.outputdir = this.options.outputdir ?? process.cwd();
+    this.validateOptions();
+  }
+  public abstract validateOptions(): void;
```
xyc, line 19: whether we have a default implementation for `validateOptions` is up for debate. Currently all commands would have some validation

packages/templates/src/utils/types.ts
```diff
@@ -14,3 +14,29 @@ export interface CreateOutput {
   created: string[];
   rawOutput: string;
 }
+
+/**
+ * Available Template types
+ */
+export enum TemplateType {
+  AnayticsTemplate,
```
lcampos, line 22:
```suggestion
  AnalyticsTemplate,
```

packages/templates/src/utils/types.ts
```diff
@@ -14,3 +14,29 @@ export interface CreateOutput {
...
+  ApexClass,
+  ApexTrigger,
+  LightningApp,
+  LightningComponent,
+  LightningEvent,
+  LightningInterface,
+  LightningTest,
+  Project,
+  VisualForceComponent,
```
lcampos, line 31:
```suggestion
  VisualforceComponent,
```

packages/templates/src/utils/types.ts
```diff
@@ -14,3 +14,29 @@ export interface CreateOutput {
...
+  VisualForcePage
```
lcampos, line 32:
```suggestion
  VisualforcePage
```

packages/templates/src/service/templateService.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  /**
+   * Create using templates
+   * @param templateType template type
+   * @param templateOptions template options
+   */
+  public async create<TOptions extends TemplateOptions>(
+    templateType: TemplateType,
+    templateOptions: TOptions
```
brpowell, line 53: No big deal for this PR but for curiosity's sake - can we explore something like the following toy example to get type checking on the options based on the template you select?

Screen Shot 2020-08-05 at 3 37 11 PM

xyc: Yeah, we could export the types for it and use `await templateService.create<ApexTriggerOptions>(TemplateType.ApexTrigger, ...)`. See the design doc linked

xyc: Oh I see what you mean. (My prev comment is about current design) In that case available template options seem to be need to be centrally defined, but this is worth looking into

brpowell: Cool cool. Does that method require someone to explicitly set the options type to use?

xyc: You don't need to explicitly set the option types unless you want to auto-complete template specific options.

I tried the above approach. All options are available though. If we want to limit to specific options, we'd need to set the type for `create`.
Screen Shot 2020-08-05 at 5 12 43 PM

Also, it might be too strict for CLI usage since flags are untyped
Screen Shot 2020-08-05 at 5 15 03 PM

xyc: Forgot to set `templateType` to also use the generic. So type completion is working as expected with below. 

```js
import { ApexClassOptions } from '../generators/apexClassGenerator';
import { LightningTestOptions } from '../generators/lightningTestGenerator';

type AllTemplateOptions = {
  [TemplateType.ApexClass]: ApexClassOptions;
  [TemplateType.LightningTest]: LightningTestOptions;
};

public async create<T extends keyof AllTemplateOptions>(
    templateType: T,
    templateOptions: AllTemplateOptions[T]
  ): Promise<CreateOutput>;
```

xyc: I had to use `T extends keyof AllTemplateOptions` instead of `T extends TemplateType`, but otherwise works fine. This doesn't actually change the function signature, but it would be very nice to have

COMMANDS.md
```diff
@@ -289,7 +289,7 @@ OPTIONS
 
   -s, --namespace=namespace                                                         project associated namespace
 
-  -t, --template=standard|empty                                                     [default: standard] template to use
+  -t, --template=standard|empty|analytics                                           [default: standard] template to use
```
xyc, line 292: this is missing `anaytics` template type

packages/templates/src/service/templateService.ts
xyc: I'm holding off on this change since this will also add overhead for adding new templates and maintaining templates, but this is the commit: https://github.com/forcedotcom/salesforcedx-templates/commit/e755014fd879fad4ad750325a8875706e66235a1

packages/templates/package.json
```diff
@@ -18,17 +18,31 @@
...
-    "typescript": "^3.8.3"
+    "typescript": "^3.8.3",
+    "yeoman-assert": "^3.1.1"
   },
   "scripts": {
     "clean:lib": "shx rm -rf lib && shx rm -rf coverage && shx rm -rf .nyc_output",
     "compile": "tsc -b",
     "build": "yarn run clean:lib && yarn build:templates && yarn compile",
     "build:templates": "node scripts/build-templates",
+    "test": "nyc --exclude src/i18n/localization.ts --extension .ts mocha --forbid-only \"test/**/*.test.ts\"",
```
xyc, line 45: excluding `localization.ts` for coverage here - until it's refactored into npm / `@salesforce/core`

packages/templates/test/mocha.opts
```diff
@@ -0,0 +1,5 @@
+--require ts-node/register
```
lcampos, line 1: Not for this PR, eventually we should move to compiling the tests using typescript instead of ts-node like we are now doing on SDR and Apex library cc @brpowell

xyc: Sounds good. Currently TS compilation errors in tests do seem to surface locally

packages/templates/test/service/templateService.test.ts
```diff
@@ -0,0 +1,121 @@
...
+          'LibraryCreateClass.cls'
+        ),
+        path.join(
+          'testsoutput',
+          'libraryCreate',
+          'apexClass',
+          'LibraryCreateClass.cls-meta.xml'
+        )
+      ]);
+      assert.fileContent(
```
lcampos, line 58: We should also assert the content of the `*.cls-meta.xml` file.

packages/templates/test/utils/createUtil.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+import { assert, expect } from 'chai';
+import * as fs from 'fs';
+import * as path from 'path';
+import { SinonStub, stub } from 'sinon';
+
+/* tslint:disable: no-unused-expression */
+describe('CreateUtil', () => {
+  describe('checkInputs', () => {
+    it('should throw error for input with non-alphanumeric characters', () => {
```
lcampos, line 18: Do we need to check for empty inputs ?

xyc: yeah it should throw an AlphaNumericNameError error too. I'll add that test case.

packages/templates/test/service/templateService.test.ts
xyc: Cool added `*.cls-meta.xml` assertion too.

PR #173, opened 2020-08-13
Publish npm packages

xyc:
### What does this PR do?
- CircleCI scripts for publishing library and plugin
- Update package versions to 49.4.0
- Rename `salesforcedx-templates` to `@salesforce/plugin-templates`

@W-7746999@


.circleci/config.yml
```diff
@@ -81,6 +125,51 @@ jobs:
           command: |
             npx codecov --disable=gcov
 
+  prepublish:
+    description: 'Prepublish'
+    working_directory: ~/cli
+    docker:
```
AnanyaJha, line 131: I don't think we *have* to specify a working directory for this job

.circleci/config.yml
```diff
@@ -81,6 +129,48 @@ jobs:
...
+            - ./packages/plugin-templates/oclif.manifest.json
+
+  publish-library:
+    description: 'Publish library package'
+    docker:
+      - image: node:12
+    working_directory: ~/cli/packages/templates
+    steps:
+      - npm-release-management/publish:
+          use_tarfile: true
```
AnanyaJha, line 159: Do we need the git tag step for the library publish?

.circleci/config.yml
xyc: If we don't specify it would just default to `~/project`. I'm keeping it consistent with current jobs

.circleci/config.yml
xyc: Nope, git tag is on the repo level

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "salesforcedx-templates-repo",
   "private": true,
   "description": "Salesforce CLI scaffolding commands for different types of Force.com metadata",
-  "version": "49.0.0",
+  "version": "49.3.0",
```
AnanyaJha, line 5: should we bump this version up to `49.4.0` as well?

xyc: yup. this version shouldn't matter, but I'll update it

xyc: Actually let me just fix it to `49.0.0`. This package wouldn't be published and we don't need to bump the version here

.circleci/config.yml
AnanyaJha: where do we push the new tag that we generated at the repo level?

xyc: https://github.com/forcedotcom/salesforcedx-templates/pull/173/files#diff-1d37e48f9ceff6d8030570cd36286a61R171-R172

AnanyaJha: thanks! I misunderstood, thought we needed to push the tag for both packages

scripts/publish-workflow.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
```
lcampos, line 2: Are we not adding the changes needed to trigger this from a VSCode task ?

.circleci/config.yml
```diff
@@ -3,6 +3,54 @@ version: 2.1
...
+            git config user.name "Release Bot"
+      - checkout
+      - run:
+          name: Bump package version
+          command: |
+            npx lerna version << pipeline.parameters.publish-type >> --force-publish --no-git-tag-version --exact --yes
+            git add .
+            export RELEASE_TAG="$(node -pe "require('./lerna.json').version")"
+            git commit -m "Updated version $RELEASE_TAG"
+      - restore_cache:
```
lcampos, line 34: This is the same as the cache for `node-latest` job, we can generalize this and reference it in both places.

scripts/publish-workflow.sh
xyc: We should, let me add it

.circleci/config.yml
```diff
@@ -81,6 +129,48 @@ jobs:
...
+            - ./packages/templates/*.tgz
+            - ./packages/plugin-templates/package.json
+            - ./packages/plugin-templates/README.md
+            - ./packages/plugin-templates/yarn.lock
+            - ./packages/plugin-templates/messages
+            - ./packages/plugin-templates/lib
+            - ./packages/plugin-templates/bin
+            - ./packages/plugin-templates/oclif.manifest.json
+
+  publish-library:
```
lcampos, line 152: For new listings, does the orb default to publishing it as a public module ? We've had to include the module being public either on the `package.json` or on the publish command since `npm` defaults publishing new modules as private.

xyc: Yes, access is default to public when it comes to publish in the orb https://circleci.com/orbs/registry/orb/salesforce/npm-release-management?version=2.0.0#commands-publish

I haven't tried out this latest version, but it should publish as a public module

.circleci/config.yml
xyc: I could use an anchor on this, but I think the benefits are limited if later we orbify this. `save_cache` couldn't use the anchor either

lcampos: Got it 👍

PR #183, opened 2020-08-18
Fix persist workspace config

xyc:
### What does this PR do?
Fix persist workspace config

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -140,14 +140,15 @@ jobs:
           root: .
           paths:
             - .git
-            - ./packages/templates/*.tgz
-            - ./packages/plugin-templates/package.json
-            - ./packages/plugin-templates/README.md
-            - ./packages/plugin-templates/yarn.lock
-            - ./packages/plugin-templates/messages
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -140,21 +140,24 @@ jobs:
...
             - .git
-            - ./packages/templates/*.tgz
-            - ./packages/plugin-templates/package.json
-            - ./packages/plugin-templates/README.md
-            - ./packages/plugin-templates/yarn.lock
-            - ./packages/plugin-templates/messages
-            - ./packages/plugin-templates/lib
-            - ./packages/plugin-templates/bin
-            - ./packages/plugin-templates/oclif.manifest.json
+            - packages/templates/package.json
```
xyc, line 143: the orb publish scripts would need `package.json` too since it looks up the tgz from `package.json` version

.circleci/config.yml
```diff
@@ -140,21 +140,24 @@ jobs:
...
+            - packages/templates/*.tgz
```
xyc, line 144: removing `./` to be consistent with circleci docs

PR #187, opened 2020-08-18
ssh config fix

xyc:
### What does this PR do?
Fix ssh config for deploy, add a task to publish custom version.


.vscode/tasks.json
```diff
@@ -72,6 +78,11 @@
       "description": "Type of version to publish",
       "options": ["minor", "patch", "major"],
       "default": "minor"
+    },
```
lcampos, line 81: This won't work unless you make more changes to the shell script and the circleci config

xyc: synced with @lcampos offline, this should work, but since this serves the need only for a few times, the task might not be necessary. Going to remove it

PR #186, opened 2020-08-18
Change default eslint rules to reduce warnings on new code

ntotten:
### What does this PR do?
Disabled several eslint rules to reduce the noise on common aura code including default samples.

### What issues does this PR fix or reference?
@W-7982322@

PR #199, opened 2020-08-25
Port #186 to develop

xyc:
### What does this PR do?
Port #186 (Before monorepo refactor) to develop


xyc:
@lcampos has updated the default branch to `main`. `main` and `develop` are the branches we work on. We no longer use the `master` branch cc @ntotten

PR #200, opened 2020-08-27
Update dev scripts

lcampos:
### What does this PR do?
- Fix lint issues
- Enforce lint in CircleCI test jobs
- Add volta configuration
- Add oclif plugin link/unlink scripts on top level package.json

@W-8017763@


packages/plugin-templates/yarn.lock
```diff
@@ -435,6 +435,16 @@
   resolved "https://registry.yarnpkg.com/@salesforce/schemas/-/schemas-1.0.1.tgz#d1db56759d2b22a7688e1821aec564e979237ad2"
   integrity sha512-78pP1GB/DbIS8nSWGL0GpQ27g02drrEo0vzYdRipGYAIXHMzlh1gqEsq0pOiIQlPm1MxWyEqbmf4GG5qSVsd0Q==
 
+"@salesforce/templates@49.4.4":
```
xyc, line 438: With this lock file change we are going to install library from npm, and local changes to library might not get reflected.

lcampos: Good catch, here's a PR to remove this https://github.com/forcedotcom/salesforcedx-templates/pull/201

PR #201, opened 2020-08-27
Remove @salesforce/templates

lcampos:
### What does this PR do?
Remove @salesforce/templates from yarn.lock to prevent getting the code from npm when developing the repo

@W-8017763@


brpowell:
Won't this be added to the lockfile any time we do yarn install? Shouldn't lerna or the linking script be able to handle if you want to develop against the local version of the package?

xyc:
@brpowell this won't be added to lock file after install, only after `sfdx plugins:link`. And yes lerna handles the dependencies so we don't need to have it in the lock file

PR #205, opened 2020-09-01
Lf/new functions template

LisandroFernandezSF:
### What does this PR do?
This PR is adding a new project template which copies the same structure than 'empty', but includes a 'functions' folder  

W-7958785

### DIFF
--- a/packages/plugin-templates/src/commands/force/project/create.ts
+++ b/packages/plugin-templates/src/commands/force/project/create.ts
@@ -35,7 +35,7 @@ export default class Project extends TemplateCommand {
       description: MessageUtil.get('ProjectTemplateFlagDescription'),
       longDescription: MessageUtil.get('ProjectTemplateFlagLongDescription'),
       default: 'standard',
-      options: ['standard', 'empty', 'analytics']
+      options: ['standard', 'empty', 'analytics', 'functions']
     }),
     outputdir: flags.string({
       char: 'd',
@@ -70,6 +70,7 @@ export default class Project extends TemplateCommand {
   };
   public async run(): Promise<AnyJson> {
     // namespace is a reserved keyword for the generator
+
     this.flags.ns = this.flags.namespace;
 
     return this.runGenerator(ProjectGenerator);
--- a/packages/plugin-templates/src/commands/force/project/create.ts
+++ b/packages/plugin-templates/src/commands/force/project/create.ts
@@ -35,7 +35,7 @@ export default class Project extends TemplateCommand {
       description: MessageUtil.get('ProjectTemplateFlagDescription'),
       longDescription: MessageUtil.get('ProjectTemplateFlagLongDescription'),
       default: 'standard',
-      options: ['standard', 'empty', 'analytics', 'functions']
+      options: ['standard', 'empty', 'analytics']
     }),
     outputdir: flags.string({
       char: 'd',
--- a/packages/templates/src/generators/projectGenerator.ts
+++ b/packages/templates/src/generators/projectGenerator.ts
@@ -205,7 +205,6 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
       }
     }
 
-    // tslint:disable-next-line:no-unused-expression
     if (template === 'functions') {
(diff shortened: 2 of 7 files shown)


packages/templates/src/templates/project/functions/README.md
```diff
@@ -0,0 +1,9 @@
+# Salesforce App
```
xyc, line 1: might be helpful if we have some get started info for users here. @smaddox-sf

packages/templates/src/utils/types.ts
```diff
@@ -107,12 +107,13 @@ export interface LightningTestOptions extends TemplateOptions {
 export interface ProjectOptions extends TemplateOptions {
   projectname: string;
   defaultpackagedir: string;
+  defaultfunctionsdir: string;
```
xyc, line 110: Where is this used?

packages/templates/src/generators/projectGenerator.ts
```diff
@@ -204,6 +204,18 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
         );
       }
     }
+
+    // tslint:disable-next-line:no-unused-expression
+    if (template === 'functions') {
```
AnanyaJha, line 209: do we need the tslint disable here? it seems like it runs fine without it

LisandroFernandezSF: Done!

packages/templates/src/utils/types.ts
LisandroFernandezSF: It's not. It was part of another approach. Fixed!

packages/templates/src/templates/project/functions/Manifest.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<Package xmlns="http://soap.sforce.com/2006/04/metadata">
+    <types>
+        <members>*</members>
+        <name>ApexClass</name>
+    </types>
+    <types>
```
smithgp, line 7: Do functions have a metadata type? If so, it would be nice to include it here so that it works right away for users.

lcampos: They do but scaffolding that is still a decision to be made and won't be addressed as part of this PR.

packages/templates/src/templates/project/functions/README.md
lcampos: That's a good suggestion that should be addressed as a separate work item for now.

packages/templates/src/templates/project/functions/ScratchDef.json
```diff
@@ -0,0 +1,5 @@
+{
+  "orgName": "<%= company %>",
+  "edition": "Developer",
+  "features": []
```
lcampos, line 4: I think functions has it's own scratch org feature that we should add here.Let me confirm the name for it.

lcampos, line 4:
```suggestion
  "features": ["Functions"]
```

packages/plugin-templates/src/commands/force/project/create.ts
```diff
@@ -35,7 +35,7 @@ export default class Project extends TemplateCommand {
       description: MessageUtil.get('ProjectTemplateFlagDescription'),
       longDescription: MessageUtil.get('ProjectTemplateFlagLongDescription'),
       default: 'standard',
-      options: ['standard', 'empty', 'analytics']
+      options: ['standard', 'empty', 'analytics', 'functions']
```
lcampos, line 38: This is correct but I wonder we might want to keep the functionality strictly to the library for now so that it's not surfaced through the cli command options. What do you think @smaddox-sf ?

lcampos: Circling back from our conversation during standup, we need to remove the changes on this PR that are on the `plugin-templates` module. The rest can be merged cc @LisandroFernandezSF

lcampos (approved):

xyc (approved):

PR #220, opened 2020-10-01
chore: add ability to provide custom version to publish (@W-8148526)

sfsholden:
### What does this PR do?
In order to ensure that we have the same version as VSCode Extensions, provide the ability to specify a version to publish.

@W-8148526@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,9 @@ parameters:
   publish-type:
     type: string
     default: minor
+  publish-version:
+    type: string
+    default: ''
 
 commands:
   setup-publish:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -40,7 +43,17 @@ commands:
       - run:
           name: Bump package version
           command: |
-            npx lerna version << pipeline.parameters.publish-type >> --force-publish --no-git-tag-version --exact --yes
+            if [ ! -z << pipeline.parameters.publish-version >> ]; then
+              if [[ << pipeline.parameters.publish-version >> =~ ^[0-9]{2}\.[0-9]{1,2}\.[0-9]{1,2}$ ]]; then
```
sfsholden, line 47: Future enhancement: Be able to pull the version from the main branch for vscode extensions.

scripts/publish-workflow.sh
```diff
@@ -5,11 +5,23 @@
 
 CircleCIToken=$1
 PublishType=$2
+PublishVersion=$3
+
+if [ ! -z "${PublishVersion}" ]; then
```
sfsholden, line 10: Adding this logic here as well so a user does not have to wait for jobs to run to know that their version is valid. The checks within the config.yml file are still required though - in the event that someone sends the API request through a different touch point.

.circleci/config.yml
```diff
@@ -15,6 +15,9 @@ parameters:
   publish-type:
     type: string
     default: minor
+  publish-version:
```
xyc, line 18: I felt we could just keep one pipeline param here: `publish-version`. (both are params to `lerna version`)

sfsholden: Nice suggestion, I'll work on changing this on Monday. I'm going to keep the version verification, and will add a check for 'patch', 'minor' or 'major' as well.

.circleci/config.yml
sfsholden: Additional TODO prior to merge: verify the version is not older than the current one in main.

sfsholden:
Closing - this is not a feature we want to add in for templates.

PR #225, opened 2020-10-09
Add query-results dir to gitignore

jonnyhork:
### What does this PR do?
This PR will add a new `query-results` directory to the `.gitignore` of the project template.
As part of the new soql query builder extension, customers will be able to save results from a soql query to the disk however we don't anticipate they will want to check in those results to source control by default.
@W-8136966@

### DIFF
--- a/packages/templates/src/templates/project/gitignore
+++ b/packages/templates/src/templates/project/gitignore
@@ -12,6 +12,9 @@
 # LWC Jest coverage reports
 coverage/
 
+# SOQL Query Results
+**/scripts/soql/query-results
+
 # Logs
 logs
 *.log
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,9 @@ orbs:
   npm-release-management: salesforce/npm-release-management@2.0.0
 
 parameters:
+  fingerprint:
+    type: string
+    default: '1e:57:e5:df:0e:82:aa:ff:33:b7:8c:0f:9c:28:56:f3'
   publish:
     type: boolean
     default: false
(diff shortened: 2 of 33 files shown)


packages/templates/src/templates/project/gitignore
```diff
@@ -12,6 +12,9 @@
 # LWC Jest coverage reports
 coverage/
 
+# SOQL Query Results
+**/scripts/soql/query-results
```
jonnyhork, line 16: I chose to be more explicit with the target directory incase a customer wants to create their own `query-results` dir that is checked in, we don't exclude that name from being used everywhere.

jonnyhork:
From what I can tell the `scripts` directory is not included in when pushing source to an org, so I did not change the .forceignore but please correct me if I am missing something.

mysticflute (approved):

packages/templates/src/templates/project/gitignore
brpowell: Are results automatically saved to this path?

brpowell:
Can you change the base branch to target `develop` instead of `main`?

@lcampos we might need to update the default branch on this repo to develop.

lcampos:
> Can you change the base branch to target `develop` instead of `main`?
> 
> @lcampos we might need to update the default branch on this repo to develop.

@brpowell done

packages/templates/src/templates/project/gitignore
jonnyhork: yes results are automatically saved to this path.

jonnyhork (commented):
@brpowell 
Sorry about the mix up with main, I changed the target branch to develop and pulled that into my feature branch but there still seems to be some weirdness going on.

It might be easier for me to just make a new PR unless the changes you see here are ok?

brpowell:
@jonnyhork yeah because the commits on main are technically different you'll have to make a new branch based off of develop, sorry for the mix up. You should be able to cherry-pick the commit from this branch and open a new PR.

brpowell:
@lcampos we might also need to add Jonny as a contributor so the builds run

jonnyhork:
New PR at https://github.com/forcedotcom/salesforcedx-templates/pull/227

PR #229, opened 2020-10-12
chore: update repo for api version 50.0

lcampos:
### What does this PR do?
Updates the repository to use v50.0 as the default api version

@W-8205897@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ commands:
             npx lerna version << pipeline.parameters.publish-type >> --force-publish --no-git-tag-version --exact --yes
             git add .
             export RELEASE_TAG="$(node -pe "require('./lerna.json').version")"
-            git commit -m "Updated version $RELEASE_TAG"
+            git commit -m "chore: updated version $RELEASE_TAG"
       - run:
           name: Build project
           command: yarn build
--- a/packages/plugin-templates/test/commands/force/analytics/template/create.test.ts
+++ b/packages/plugin-templates/test/commands/force/analytics/template/create.test.ts
@@ -29,7 +29,7 @@ describe('Analytics template creation tests:', () => {
             path.join('waveTemplates', 'foo', 'template-info.json'),
             {
               label: 'foo',
-              assetVersion: 49.0
+              assetVersion: 50.0
             }
           );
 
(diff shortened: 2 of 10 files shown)


.circleci/config.yml
```diff
@@ -43,7 +43,7 @@ commands:
             npx lerna version << pipeline.parameters.publish-type >> --force-publish --no-git-tag-version --exact --yes
             git add .
             export RELEASE_TAG="$(node -pe "require('./lerna.json').version")"
-            git commit -m "Updated version $RELEASE_TAG"
+            git commit -m "chore: updated version $RELEASE_TAG"
```
lcampos, line 46: @sfsholden updated the circleci commits to follow the new commitizen format

packages/plugin-templates/test/commands/force/analytics/template/create.test.ts
```diff
@@ -29,7 +29,7 @@ describe('Analytics template creation tests:', () => {
             path.join('waveTemplates', 'foo', 'template-info.json'),
             {
               label: 'foo',
-              assetVersion: 49.0
+              assetVersion: 50.0
```
lcampos, line 32: At some point we should update the tests to read the version from `package.json`, but that will be on a different PR.

packages/plugin-templates/test/commands/force/analytics/template/create.test.ts
```diff
@@ -29,7 +29,7 @@ describe('Analytics template creation tests:', () => {
             path.join('waveTemplates', 'foo', 'template-info.json'),
             {
               label: 'foo',
-              assetVersion: 49.0
```
smithgp, line 32: This should not be updated, nor the assetVersion in the template-info.json below. This number refers not the current version of the server, but to the format of the dashboard.json file, which is 49.0 -- the analytics server code uses this value to determine the parser & upgraders to use at runtime.
In analytics, It's normal to leave this assetVersion as-is until the format actually changes.

smithgp (changes requested):

packages/plugin-templates/test/commands/force/analytics/template/create.test.ts
lcampos: Cool, I've reverted this

smithgp (approved):

AnanyaJha (approved):

xyc (approved):

PR #230, opened 2020-10-12
JHork: Add query-results to .gitignore

jonnyhork:
### What does this PR do?
This PR will add a new query-results directory to the .gitignore of the project template.
As part of the new soql query builder extension, customers will be able to save results from a soql query to the disk however we don't anticipate they will want to check in those results to source control by default.

@W-8136966@

### DIFF
--- a/packages/templates/src/templates/project/gitignore
+++ b/packages/templates/src/templates/project/gitignore
@@ -12,6 +12,9 @@
 # LWC Jest coverage reports
 coverage/
 
+# SOQL Query Results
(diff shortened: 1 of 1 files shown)


jonnyhork:
@brpowell 3rd time is a charm?

brpowell:
😂  thanks for being a good sport, sorry for the confusion @jonnyhork

brpowell (approved):

jonnyhork:
All good =)

PR #240, opened 2020-11-11
fix: publish library module without using orb

lcampos:
### What does this PR do?
Update CircleCI to publish the library package without using `npm-release-management` orb

@W-8384687@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,8 +172,16 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/cli
-      - npm-release-management/publish:
-          use_tarfile: true
+      - run:
+          name: Set .npmrc
+          command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
+      -run:
+          name: Publish to NPM
+          command: |
+            RELEASE_VERSION="$(node -pe "require('./package.json').version")"
+            TARFILE=$(ls | grep $RELEASE_VERSION.tgz)
+            echo "publishing to npm using ${TARFILE}"
+            npm publish $TARFILE --access public
 
   publish-plugin:
     description: 'Publish plugin package'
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -56,4 +56,5 @@ GitIgnoreTest2/
 oclif.manifest.json
 
 # file generated for publishing
-*.tgz
\ No newline at end of file
+*.tgz
+.npmrc
```
lcampos, line 60: Added this to make sure we don't accidentally commit the `.npmrc` file

.circleci/config.yml
```diff
@@ -172,8 +172,16 @@ jobs:
...
+      - run:
+          name: Set .npmrc
+          command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
+      - run:
+          name: Publish to NPM
+          command: |
+            RELEASE_VERSION="$(node -pe "require('./package.json').version")"
+            TARFILE=$(ls | grep $RELEASE_VERSION.tgz)
+            echo "publishing to npm using ${TARFILE}"
+            npm publish $TARFILE --access public
```
brpowell, line 184: What does the orb do differently from this?

brpowell (approved):

.circleci/config.yml
lcampos: It runs a script that pings npm for the status of the package that was published. That is what is currently timing out and breaking the build.

.circleci/config.yml
```diff
@@ -172,8 +172,16 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/cli
-      - npm-release-management/publish:
-          use_tarfile: true
+      - run:
+          name: Set .npmrc
+          command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
```
sfsholden, line 177: Pointing this out in case it makes a difference:
The npm-release-management orb has `$HOME/.npmrc`

sfsholden (approved):

.circleci/config.yml
lcampos: I think that should be fine, npm will pick up the config as long as it part of the workspace. I this case I'm creating it at the top level like we do with the other jobs.

PR #289, opened 2021-03-25
Added default Jest config in new project template

pozil:
This PR adds a minimal Jest configuration that's compatible with LWC.
This allows VSCode Jest extensions and VSCode debugger to run.
I've excluded the local dev server so that it's not included in Jest code code coverage.

FYI @albarivas


brpowell:
Hi @pozil. Are you looking to add this file to the standard project:create template?

pozil:
Hi @brpowell. Yes, that's my intent.

brpowell:
Cool. You'll need to move this file to the project template folder here:

https://github.com/forcedotcom/salesforcedx-templates/tree/develop/packages/templates/src/templates/project

Then make sure the project generator copies it here:

https://github.com/forcedotcom/salesforcedx-templates/blob/3a45a0844257ec4f4d72b787527d8f933b8734db/packages/templates/src/generators/projectGenerator.ts#L167

And then validate that the file is copied correctly in the test case here:

https://github.com/forcedotcom/salesforcedx-templates/blob/develop/packages/plugin-templates/test/commands/project/create.test.ts

pozil:
Thanks for the guidance @brpowell. I applied your comments.

brpowell:
@pozil I tested project create, npm install, creating a test and running it - everything looks good 👍 . I'll run the builds so we can get this merged and released for next week.

Does it make sense to put this jest config in the lwc folder with the rest of the LWC specific configuration? (eslint, jsconfig) I'm not too familiar with valid LWC project structure, maybe it is better at the root but thought I would ask.

PR #299, opened 2021-04-13
chore: update Aura and LWC dependencies for 51.0

### LINKED ISSUE #246: Node devDependencies are out of date in new project template
Node devDependencies are out of date in new project template.
This is particularly problematic for `sfdx-lwc-jest` as the lib version does not support testing current Salesforce API version.

pmdartus:
### What does this PR do?

This PR updates the Aura and LWC related dependencies to Salesforce API version 51. As of today:
- Running `npm install` after `force:project:create --projectname mywork` produces an error because `@salesforce/eslint-plugin-aura@^1` is not compatible with `eslint@7`. This is fixed by the new `2.0.0` release of `@salesforce/eslint-plugin-aura`.
- Running `npm run test:unit` fails because `@salesforce/sfdx-lwc-jest@^0.9` is compatible with API version 50. Updating to `^0.11` fixes this issue.

Fixes #246 and W-9079074


packages/templates/src/templates/project/aura.eslintrc.json
```diff
@@ -2,7 +2,6 @@
   "plugins": ["@salesforce/eslint-plugin-aura"],
   "extends": ["plugin:@salesforce/eslint-plugin-aura/recommended", "prettier"],
   "rules": {
-    "func-names": "off",
```
pmdartus, line 5: This rule was dropped with `@salesforce/eslint-plugin-aura^2`

brpowell (approved):

ravijayaramappa (approved):

trevor-bliss:
@brpowell Any ideas for how we can have automated checks to verify the node dependencies and package.json scripts in the project templates are functional?

It'd be really valuable to know when the default templates have issues and it may not be a very common flow for the internal developers who maintain these projects. A script that creates a new project from a template and runs the format, lint, and test commands against some simple components would provide a lot of value.

brpowell:
@trevor-bliss Sounds like a good idea, I could see it as a test in the templates package. In terms of the verification build we might spin it off as its own job since it might take longer to run, but @sfsholden might have some ideas on how to handle that aspect.

Feel free to file an issue for this or open a PR with an initial take. It's not something we'll be able to look at for at least a couple months likely.

trevor-bliss:
Thanks @brpowell. I filed an issue https://github.com/forcedotcom/salesforcedx-templates/issues/308. I'd be happy to talk discuss it further.

PR #314, opened 2021-05-05
Ignore common dotenv files by default

jonathanwiesel:
### What does this PR do?
Make `.env` files part of the default `.gitignore` of the project generator template, since there are use cases where you may be using environment files and those should never be committed to VCS


brpowell:
Thanks for the contribution @jonathanwiesel!

PR #316, opened 2021-05-05
add code folders to analytics projects 

smithgp:
### What does this PR do?
In `force:project:create -t analytics`:
* Include `aura/`, `classes/`, and `lwc/` folders.
* Include the aura and lwc eslint configs.
* Adds the analytics vscode extension to the extension recommendations.

Also, updates the node version in `.nvmrc` to match the volta config entry.

@W-9191555@


reidaelliott (approved):

packages/templates/src/generators/projectGenerator.ts
```diff
@@ -124,30 +123,14 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
       this.fs.copyTpl(
         this.templatePath('lwc.eslintrc.json'),
         this.destinationPath(
-          path.join(
```
smithgp, line 127: This is just removing some duplicate code used in a number of places (`folderlayout` was already defined to point to the `default` folder path).

packages/templates/src/generators/projectGenerator.ts
```diff
@@ -175,7 +158,6 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
       }
     }
 
-    // tslint:disable-next-line:no-unused-expression
```
smithgp, line 178: These tslint disables don't appear to be needed any longer.

.nvmrc
```diff
@@ -1 +1 @@
-v10.15.3
\ No newline at end of file
+v12.4.0
```
brpowell, line 1: Were there issues with the previous node version?

smithgp: TBH, I didn't try, I just saw the mismatch from the volta config and switched it.
I just did `yarn install && yarn run build && yarn run test` with 10.15.3 and it was fine.
In any case, the versions in `.nvmrc` and `volta` in `package.json` (and whatever circleci does) should probably line up.

brpowell: Ah yeah that makes sense. At first I thought this was a project template file heh.

brpowell (approved):

Issue #98, opened 2020-03-02
Support External Repository for Project Templates


ntotten:
Support external samples repository like `create-next-app`. See: https://nextjs.org/blog/create-next-app

The idea would be we have a github repo `salesforcedx-examples` or something. Each example is in a folder `example-1`, `example-2`, etc. The user can then run the command:

```
sfdx force:project:create --template <name>
```

We could also make the example repo customizable so team's could create their own custom templates.

```
sfdx force:project:create --template <name> --repository https://github.com/myorg/templates
```

jonathanwiesel:
This would be great since many companies and projects require development environments tweaking, this will grant the required flexibility to do it.

randi274:
We now offer this feature! 🎉  Check out our documentation here for more details on implementing this for your project and your team: https://developer.salesforce.com/tools/vscode/en/user-guide/byotemplate

jclark-dot-org:
@randi274 That's great but it's not really what this request was asking for, which is the ability to create a new SFDX project from a template other than the two that ship with SFDX.  While the linked item looks useful, sfdx is a command-line tool whereas VSCode is an IDE.

randi274:
Hey @jclark-dot-org thanks for the feedback! The custom templates feature should work for both VS Code and the CLI, as they both make use of this library (so the docs referring to how to set the config variables work in both cases!).

But after I saw your comment, I ran through some additional testing of projects specifically, and I'm seeing that projects isn't quite configurable yet (either within the default 3 of `standard`, `empty`, or `analytics`, or modifying each of them individually), so it's a good case to re-open here 👍 cheers!

PR #325, opened 2021-05-19
feat: add analytics dashboard lwc templates

smithgp:
### What does this PR do?

- Fixes using `--template` with `force:lighting:component:create`:
  - `--help` didn't list any available template names
  - actually specifying `--template` with any value (including the default `DefaultLightningCmp`) would produce an error 
- Updates the `force:lightning:component:create` command to support multiple options for `--template`
- Adds `analyticsDashboard` and `analyticsDashboardWithStep` lwc templates

@W-9191596@

cc: @jimmydief


packages/templates/src/generators/lightningComponentGenerator.ts
```diff
@@ -60,63 +80,62 @@ export default class LightningComponentGenerator extends SfdxGenerator<
...
         ),
         {}
-      ),
-        this.fs.copyTpl(
-          this.templatePath(`${template}.cmp`),
-          this.destinationPath(
-            path.join(outputdir, componentname, `${componentname}.cmp`)
-          ),
-          {}
+      );
```
smithgp, line 88: These calls where all ending with a `,` (and not `;` before, which, while valid, seemed a little weird. I updated to use a `;`, which also fixed the formatting -- turn off whitespace diffs in the gear icon to see this a little better.

packages/plugin-templates/messages/messages.json
```diff
@@ -64,6 +64,7 @@
   "MissingAuraFolder": "Lightning bundles must have a parent folder named 'aura'.",
   "MissingAppname": "Missing required flag:\n -n, --appname APPNAME  name of the generated Lightning app\nSee more help with --help\n",
   "MissingComponentName": "Missing required flag:\n -n, --componentname COMPONENTNAME  name of the generated Lightning component\nSee more help with --help\n",
+  "MissingLightningComponentTemplate": "Template %s not available for component type %s",
```
jshackell-sfdc, line 67: Tiny edit:  Please add a period at the end of the message.  (while you're at it, do the same with the other messages...) thx.

smithgp: It looks like all the ones that end in `See more help with --help` are only used during tests, to match the output generated by the base oclif libraries (so they shouldn't be updated here).

smithgp: I fixed up the `MissingLightningComponentTemplate` one to have an ending `.`.

packages/templates/src/utils/types.ts
```diff
@@ -50,10 +50,10 @@ export interface AnalyticsTemplateOptions extends TemplateOptions {
 
 export interface ApexClassOptions extends TemplateOptions {
   template:
-  | 'DefaultApexClass'
-  | 'ApexUnitTest'
-  | 'ApexException'
-  | 'InboundEmailService';
+    | 'DefaultApexClass'
```
smithgp, line 53: The reformat in pre-commit did these whitespace changes here.

packages/templates/src/utils/types.ts
```diff
@@ -81,7 +81,7 @@ export interface LightningAppOptions extends TemplateOptions {
 
 export interface LightningComponentOptions extends TemplateOptions {
   componentname: string;
-  template: 'DefaultLightningCmp';
+  template: string;
```
xyc, line 84: can we use a union type here?

smithgp: Yes, sure, although it would be disconnected from the valid values in the `options` on the flag config for `--template` and need to be updated manually if new templates are added (like it was before my change here).

xyc: I'm not sure `getCommandTemplatesInSubdirs` is the abstraction we need at the moment. For now template libs only support existing *static* templates defined in folders. From an API perspective we don't allow arbitrary templates nor user provided templates, so this can't be a generic string.

For more flexible template creation, for example [Support External Repository for Project Templates](https://github.com/forcedotcom/salesforcedx-templates/issues/98) the template of the project options indeed needs to be a generic `string` , however that's a different topic (no plan to support it yet) and requires more thinking on the API and implementations, which I don't think applies here.

smithgp: Ok, that makes sense. I was following the pattern in the existing command that was using `getCommandTemplatesForFiletype` for the options, but I agree it's not necessary for now.

I'm going to change this to a union of these templates (`'default' | 'analyticsDashboard' | 'analyticsDashboardWithStep'`) plus update the `options` on the flag config for `--template` to also just be the hard-coded list for consistency.

smithgp: Done.

xyc: Thanks, In an ideal world, we could probably have a better option than `getCommandTemplatesForFiletype` too. This type should be statically generated (instead of dynamically interpreted).

Also don’t worry too much about the hard-coding / duplication. Check out this talk [minimal API surface area](https://2014.jsconf.eu/speakers/sebastian-markbage-minimal-api-surface-area-learning-patterns-instead-of-frameworks.html) which makes a fair point about that. 

> And, that means that it can actually be more ‑‑ a lot easier to upgrade verbose, repetitive and explicit code than it is to upgrade an abstraction.

xyc (approved):

PR #334, opened 2021-06-02
bump @salesforce/core to2.23.2

xyc:
### What does this PR do?
bump @salesforce/core to 2.23.2

### DIFF
--- a/packages/plugin-templates/package.json
+++ b/packages/plugin-templates/package.json
@@ -101,9 +101,6 @@
     }
   },
   "repository": "forcedotcom/salesforcedx-templates",
-  "resolutions": {
-    "**/@salesforce/core": "2.23.2"
-  },
   "scripts": {
     "clean:lib": "shx rm -rf lib && shx rm -rf coverage && shx rm -rf .nyc_output && shx rm -f oclif.manifest.json",
     "compile": "tsc -b",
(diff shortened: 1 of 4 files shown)


packages/plugin-templates/package.json
```diff
@@ -101,6 +101,9 @@
     }
   },
   "repository": "forcedotcom/salesforcedx-templates",
+  "resolutions": {
+    "**/@salesforce/core": "2.23.2"
```
xyc, line 105: Adding a resolution field to fix the resolved version, similar to apex.

xyc:
The integration tests failures are test issues only. This commit https://github.com/forcedotcom/sfdx-core/commit/f5cfe52605e54ab8250e155e17077bc03b0cd6ec stubs `resolveProjectPath ` before each test. [withProject](https://github.com/forcedotcom/cli-packages/blob/b089ff3508131aa35774ee088376b7a98dcb3167/packages/command/src/test/index.ts#L111) should have some handling of the already stubbed case, for example restoring before stubbing.

xyc:
Reverting the fixed resolution field for now. It only applies to plugin-templates. It's fine for now to have a mixed version of core lib with regards to `@salesforce/command`. We can re-apply the resolution field after https://github.com/forcedotcom/cli-packages/pull/155 goes in and updating `@salesforce/command` to latest version.

jag-sfdc (approved):

PR #351, opened 2021-06-24
Port v52.1.0 to develop

xyc:
### What does this PR do?
Port to develop v52.1.0

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "52.0.0",
+  "version": "52.1.0",
   "npmClient": "yarn",
   "packages": ["packages/*"]
 }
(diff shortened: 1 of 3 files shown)


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "52.0.0",
+  "version": "52.1.0",
```
rcoringrato-sfdc, line 2: @xyc Is the template release 1 version back from extensions?

xyc: Generally the lib versions don't have a correlation with extension version since they are not guaranteed to be published every week. We'll want the major version here to be consistent though.

AnanyaJha (approved):

rcoringrato-sfdc (approved):

PR #367, opened 2021-09-10
Implement custom templates in template library

xyc:
@W-9790296@

### DIFF
--- a/packages/templates/src/generators/sfdxGenerator.ts
+++ b/packages/templates/src/generators/sfdxGenerator.ts
@@ -4,6 +4,9 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+
+import * as fs from 'fs';
+import * as path from 'path';
 import * as Generator from 'yeoman-generator';
 import { TemplateService } from '../service/templateService';
 import { TemplateOptions } from '../utils/types';
@@ -18,6 +21,12 @@ export abstract class SfdxGenerator<
     apiversion: string;
     outputdir: string;
   };
+
+  /**
+   * Set by sourceRootWithPartialPath called in generator
+   */
+  public builtInTemplatesRootPath?: string;
+
   constructor(args: string | string[], options: TOptions) {
     super(args, options);
     this.options.apiversion =
@@ -29,4 +38,40 @@ export abstract class SfdxGenerator<
    * Validate provided options
    */
   public abstract validateOptions(): void;
+
+  /**
+   * Set source root to built-in templates or custom templates root if available.
+   * @param partialPath the relative path from the templates folder to templates root folder.
+   */
+  public sourceRootWithPartialPath(partialPath: string): void {
+    this.builtInTemplatesRootPath = path.join(
+      __dirname,
+      '..',
+      'templates',
+      partialPath
+    );
+    const { customTemplatesRootPath } = TemplateService.getInstance();
+    if (!customTemplatesRootPath) {
+      this.sourceRoot(path.join(this.builtInTemplatesRootPath));
+    } else {
+      if (fs.existsSync(path.join(customTemplatesRootPath, partialPath))) {
+        this.sourceRoot(path.join(customTemplatesRootPath, partialPath));
+      }
+    }
+  }
+
+  public templatePath(...paths: string[]): string {
+    // The template paths are relative to the generator's source root
+    // If we have set a custom template root, the source root should have already been set.
+    // Otherwise we'll fallback to the built-in templates
+    const customPath = super.templatePath(...paths);
+    if (fs.existsSync(customPath)) {
+      return customPath;
+    } else {
+      // files that are builtin and not in the custom template folder
+      return super.templatePath(
+        path.join(this.builtInTemplatesRootPath!, ...paths)
+      );
+    }
+  }
 }
--- a/packages/templates/src/i18n/i18n.ts
+++ b/packages/templates/src/i18n/i18n.ts
@@ -29,6 +29,13 @@ export const messages = {
   MissingLightningComponentTemplate:
     'Template %s not available for component type %s.',
 
+  localCustomTemplateDoNotExist:
+    'Local custom templates folder %s do not exist',
+  customTemplatesShouldUseHttpsProtocol:
+    'Only HTTPS protocol is supported for custom templates. Got %s.',
+  customTemplatesSupportsGitHubOnly:
+    'Unsupported custom templates repo %s. Only GitHub is supported.',
+
   LightningAppBundle: 'A Lightning Application Bundle',
   LightningComponentBundle: 'A Lightning Component Bundle',
   LightningEventBundle: 'A Lightning Event Bundle',
--- a/packages/templates/src/i18n/i18n.ts
+++ b/packages/templates/src/i18n/i18n.ts
@@ -35,6 +35,9 @@ export const messages = {
     'Only HTTPS protocol is supported for custom templates. Got %s.',
   customTemplatesSupportsGitHubOnly:
     'Unsupported custom templates repo %s. Only GitHub is supported.',
+  customTemplatesInvalidRepoUrl: 'Invalid custom templates repository URL: %s',
+  customTemplatesCannotRetrieveDefaultBranch:
+    'Cannot retrieve default branch for custom templates repository: %s',
 
   LightningAppBundle: 'A Lightning Application Bundle',
   LightningComponentBundle: 'A Lightning Component Bundle',
--- a/packages/templates/src/i18n/i18n.ts
+++ b/packages/templates/src/i18n/i18n.ts
@@ -30,7 +30,7 @@ export const messages = {
     'Template %s not available for component type %s.',
 
   localCustomTemplateDoNotExist:
-    'Local custom templates folder %s do not exist',
+    'Local custom templates folder %s does not exist',
   customTemplatesShouldUseHttpsProtocol:
     'Only HTTPS protocol is supported for custom templates. Got %s.',
   customTemplatesSupportsGitHubOnly:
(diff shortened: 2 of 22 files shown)


packages/templates/src/i18n/i18n.ts
```diff
@@ -29,6 +29,16 @@ export const messages = {
   MissingLightningComponentTemplate:
     'Template %s not available for component type %s.',
 
+  localCustomTemplateDoNotExist:
```
xyc, line 32: @sbudhirajadoc Here are the error messages for accessing a custom templates GitHub repo. Let me know if there's any questions / suggestions. Thanks!

sbudhirajadoc: Would adding the word, "custom" so "Custom template %s not available for component type %s." be additionally helpful? Also, can they take any action when they see it's missing? @xyc

xyc: Did you mean line 26-30? So those are not specific to custom templates - they can be emitted if not using a custom template, and this PR doesn't make changes to those messages.

sbudhirajadoc: no, line 30. Is that not the correct line to review? " 'Template %s not available for component type %s.'," ?

xyc: yeah those are not specific to this PR. (You could see that since the line doesn't have a `+` prepending it). The green lines are the newly added messages. Not much we can do about it though, it's GitHub's mechanism to include lines in context into review.

sbudhirajadoc (approved):

packages/templates/src/generators/sfdxGenerator.ts
```diff
@@ -29,4 +38,40 @@ export abstract class SfdxGenerator<
...
+   */
+  public sourceRootWithPartialPath(partialPath: string): void {
+    this.builtInTemplatesRootPath = path.join(
+      __dirname,
+      '..',
+      'templates',
+      partialPath
+    );
+    const { customTemplatesRootPath } = TemplateService.getInstance();
+    if (!customTemplatesRootPath) {
```
jag-sfdc, line 54: When will customTemplatesRootPath be non truthy? TemplateService.getInstance() seems to create and return a singleton if null.

packages/templates/test/service/templateService.test.ts
```diff
@@ -70,6 +79,317 @@ describe('TemplateService', () => {
...
+        return fs.existsSync(fsPath);
+      });
+      const streamStub = spy(got, 'stream');
+      const templateService = TemplateService.getInstance(process.cwd());
+      if (fs.existsSync(TEST_CUSTOM_TEMPLATES_STORAGE_PATH)) {
+        fs.removeSync(TEST_CUSTOM_TEMPLATES_STORAGE_PATH);
+      }
+      const customTemplates = TEST_CUSTOM_TEMPLATES_REPO;
+
+      const forceLoadingRemoteRepo = true;
```
randi274, line 329: Could this be false here, or even test the default false value setting? Since the folder shouldn't exist, it seems like we shouldn't need this to force the reload.

packages/templates/src/generators/sfdxGenerator.ts
xyc: If the custom templates config is not set it will be undefined

packages/templates/src/generators/sfdxGenerator.ts
```diff
@@ -29,4 +38,40 @@ export abstract class SfdxGenerator<
...
+    // The template paths are relative to the generator's source root
+    // If we have set a custom template root, the source root should have already been set.
+    // Otherwise we'll fallback to the built-in templates
+    const customPath = super.templatePath(...paths);
+    if (fs.existsSync(customPath)) {
+      return customPath;
+    } else {
+      // files that are builtin and not in the custom template folder
+      return super.templatePath(
+        path.join(this.builtInTemplatesRootPath!, ...paths)
```
randi274, line 73: Is the non-null assertion necessary here? Is a null builtInTemplatesRootPath something we can test for?

packages/templates/src/service/templateService.ts
```diff
@@ -65,11 +68,21 @@ export class TemplateService {
...
-    templateOptions: TOptions
+    templateOptions: TOptions,
+    customTemplatesRootPathOrGitRepo?: string
   ): Promise<CreateOutput> {
+    await this.setCustomTemplatesRootPathOrGitRepo(
+      customTemplatesRootPathOrGitRepo
+    );
+    if (customTemplatesRootPathOrGitRepo) {
+      // In VS Code, if creating using a custom template, we need to reset the yeoman environment
+      this.resetEnv();
```
randi274, line 83: What happens if we don't reset here?

packages/templates/src/service/templateService.ts
```diff
@@ -111,4 +124,56 @@ export class TemplateService {
       });
     });
   }
+
+  /**
+   * Local custom templates path set by user
+   * or set to a local cache of a git repo.
+   * Used by the generators.
+   */
+  public customTemplatesRootPath?: string;
```
randi274, line 133: To follow-up with @jag-sfdc's question, should we move this to the top of the file near the `getInstance` for better visibility?

packages/templates/src/service/templateService.ts
```diff
@@ -111,4 +124,56 @@ export class TemplateService {
...
+   * Throws an error if local path doesn't exist or cannot reach git repo.
+   * @param customTemplatesRootPathOrGitRepo custom templates root path or git repo
+   * @param forceLoadingRemoteRepo by default do not reload remote repo if the repo is already downloaded
+   */
+  public async setCustomTemplatesRootPathOrGitRepo(
+    pathOrRepoUri?: string,
+    forceLoadingRemoteRepo: boolean = false
+  ) {
+    if (pathOrRepoUri === undefined) {
+      this.customTemplatesRootPath = undefined;
```
randi274, line 151: @jag-sfdc I think this is where we set the undefined get set that would trigger the falsey. Could we add a test for this scenario too?

randi274 (commented):
Thanks for the tag on this @xyc! I've mostly just logged clarifying questions to improve my understanding + a few test comments. You've made great progress on this epic 🙌

packages/templates/src/service/gitRepoUtils.ts
```diff
@@ -0,0 +1,129 @@
+/*---------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
xyc: we used to have [similar questions](https://github.com/forcedotcom/salesforcedx-apex/pull/147/files#r589880268) about whether to include two licenses, but decided to keep one only.

packages/templates/src/generators/sfdxGenerator.ts
xyc: we could throw an error about builtInTemplatesRootPath being null, but since that should always be set by the generator it shouldn't happen.

packages/templates/src/service/gitRepoUtils.ts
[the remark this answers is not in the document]
randi274: Got it, thanks for the add'l context!

packages/templates/src/service/templateService.ts
xyc: The non-custom templates test should have already covered this code path

packages/templates/src/service/templateService.ts
xyc: Not sure if that's necessary. I'm grouping the related members / methods custom templates together so that they could be easily refactored if need be. If you are viewing the code - the outline feature should give a nice overview of the structure.

packages/templates/test/service/templateService.test.ts
```diff
@@ -70,6 +79,317 @@ describe('TemplateService', () => {
...
+        'public with sharing class CustomLibraryCreateClass';
+      const expectedApexClassMetaPath = path.join(
+        'testsoutput',
+        'customLibraryCreate',
+        'apexClass',
+        'LibraryCreateClass.cls-meta.xml'
+      );
+      const expectedApexClassMetaContent = `<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>52.0</apiVersion>
```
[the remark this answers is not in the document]
xyc: Good point - I was hard coding the apiVersion so that when we upgrade the major version the test will break. Should revisit the benefits that brings.

packages/templates/src/service/templateService.ts
xyc: Yeoman would reuse the template cached in memory. i.e. update to the disk files will be ignored.

packages/templates/test/service/templateService.test.ts
xyc: nice catch! will remove this

packages/templates/src/generators/sfdxGenerator.ts
jag-sfdc: Got it, this makes sense after I read through templateService 👍

packages/templates/src/service/templateService.ts
randi274: Got it, thanks for clarifying!

packages/templates/src/service/templateService.ts
randi274: Great point about developing with an eye for possible refactoring. Works for me here 👍

jag-sfdc (approved):

packages/templates/src/service/gitRepoUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+  username: string;
+  name: string;
+  branch: string;
+  filePath: string;
+}
+/**
+ * extract repo info from uri
+ * @param repoUri uri to git repo
+ */
+export async function getRepoInfo(repoUri: URL): Promise<RepoInfo> {
```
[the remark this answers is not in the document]
xyc: this is mostly adapted from create-next-app code, and it's a pretty self-contained function so I've exported it. It shouldn't affect top level exports

PR #373, opened 2021-09-27
Port pr v52.2.0

randi274:
Port PR for release v52.2.0


xyc:
Can we rebase and drop this commit: `cc28a82`? It doesn't look like it should be included

PR #376, opened 2021-09-27
Support custom templates in CLI

xyc:
### What does this PR do?
Support custom templates in CLI, utilizing the updated template library.

@W-9790334@

### DIFF
--- a/packages/plugin-templates/test/utils/templateCommand.test.ts
+++ b/packages/plugin-templates/test/utils/templateCommand.test.ts
@@ -5,9 +5,11 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { test } from '@salesforce/command/lib/test';
-import { ConfigAggregator } from '@salesforce/core';
+import { Config, ConfigAggregator } from '@salesforce/core';
 import { ForceGeneratorAdapter, Log } from '@salesforce/templates/lib/utils';
+import * as assert from 'yeoman-assert';
 
+import { nls } from '@salesforce/templates/lib/i18n';
 import { expect } from 'chai';
 import * as path from 'path';
 import { stub } from 'sinon';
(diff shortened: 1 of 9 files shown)


packages/plugin-templates/test/utils/templateCommand.test.ts
```diff
@@ -94,4 +100,135 @@ describe('TemplateCommand', () => {
...
+      .withOrg()
+      .withProject()
+      .stdout()
+      .stub(ConfigAggregator.prototype, 'getPropertyValue', (key: string) => {
+        if (key === Config.CUSTOM_ORG_METADATA_TEMPLATES) {
+          return LOCAL_CUSTOM_TEMPLATES;
+        }
+      })
+      .command(['force:apex:class:create', '--classname', 'foo'])
+      .it('should create custom template from local folder', ctx => {
```
randi274, line 146: Could we add another test to cover the scenario where there is a valid local directory, but that it doesn't contain a template for the metadata that they're creating? Not sure if it's excess, but thought to ask since we discussed that error handling in the demo yesterday.

packages/plugin-templates/test/utils/templateCommand.test.ts
```diff
@@ -94,4 +100,135 @@ describe('TemplateCommand', () => {
...
+      .withOrg()
+      .withProject()
+      .stderr()
+      .stub(ConfigAggregator.prototype, 'getPropertyValue', (key: string) => {
+        if (key === Config.CUSTOM_ORG_METADATA_TEMPLATES) {
+          return NON_EXISTENT_REPO;
+        }
+      })
+      .command(['force:apex:class:create', '--classname', 'foo'])
+      .it('should throw error if cannot retrieve default branch', ctx => {
```
randi274, line 177: Nit: could we add a more detailed description to this and the remaining tests so they're not all the same? Like `should throw error if repo doesn't exist`.

xyc: Nice catch! updated the rest of the tests. This error is more specifically about failure in making an API call to GitHub to get the default branch of the repo though (not necessarily a non-existent repo)

packages/plugin-templates/test/utils/templateCommand.test.ts
xyc: Good call! I'll add it

xyc: Added fallback tests for both local/remote

randi274 (approved):

packages/plugin-templates/test/utils/templateCommand.test.ts
randi274: Changes look great, thanks!

PR #425, opened 2022-01-14
chore: bump sfdx-core version

randi274:
### What does this PR do?
Updates sfdx-cli for security and performance purposes.

@W-8858092@

### DIFF
--- a/packages/plugin-templates/package.json
+++ b/packages/plugin-templates/package.json
@@ -9,7 +9,7 @@
     "@oclif/config": "^1",
     "@oclif/errors": "^1",
     "@salesforce/command": "^2.2.0",
-    "@salesforce/core": "2.32.0",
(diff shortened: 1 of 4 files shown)


packages/plugin-templates/package.json
```diff
@@ -9,7 +9,7 @@
     "@oclif/config": "^1",
     "@oclif/errors": "^1",
     "@salesforce/command": "^2.2.0",
```
shetzel, line 11: Can you please bump the version of `@salesforce/command` to latest as well?  It also has a dependency on `@salesforce/core` and if you leave it as is there will still be an old version of core in the dependency tree.

shetzel (approved):
👍

jag-sfdc (approved):

PR #437, opened 2022-01-25
chore: bump salesforce command

randi274:
Bump the Salesforce command library to remove dependency on sfdx-faye.

### DIFF
--- a/packages/plugin-templates/package.json
+++ b/packages/plugin-templates/package.json
@@ -8,7 +8,7 @@
     "@oclif/command": "^1",
     "@oclif/config": "^1",
     "@oclif/errors": "^1",
-    "@salesforce/command": "^2.2.0",
+    "@salesforce/command": "^4.2.1",
     "@salesforce/core": "2.33.1",
     "@salesforce/templates": "53.1.0",
     "tslib": "^1",
--- a/packages/plugin-templates/test/commands/force/analytics/template/create.test.ts
+++ b/packages/plugin-templates/test/commands/force/analytics/template/create.test.ts
@@ -13,7 +13,7 @@ describe('Analytics template creation tests:', () => {
   describe('Check analytics template creation', () => {
     test
       .withOrg()
-      .withProject()
+      //.withProject() // temporarily broken in current version of command
       .stdout()
       .command([
         'force:analytics:template:create',
@@ -64,7 +64,7 @@ describe('Analytics template creation tests:', () => {
       );
     test
       .withOrg()
-      .withProject()
+      //.withProject()
       .stderr()
       .command([
         'force:analytics:template:create',
@@ -83,7 +83,7 @@ describe('Analytics template creation tests:', () => {
       );
     test
       .withOrg()
-      .withProject()
+      //.withProject()
       .stderr()
       .command(['force:analytics:template:create'])
       .it('should throw error when missing required name field', ctx => {
@@ -91,7 +91,7 @@ describe('Analytics template creation tests:', () => {
       });
     test
       .withOrg()
-      .withProject()
+      //.withProject()
       .stderr()
       .command([
         'force:analytics:template:create',
--- a/packages/plugin-templates/test/utils/templateCommand.test.ts
+++ b/packages/plugin-templates/test/utils/templateCommand.test.ts
@@ -122,7 +122,7 @@ describe('TemplateCommand', () => {
       .withOrg()
       //.withProject()
       .stdout()
-      .stub(ConfigAggregator.prototype, 'getPropertyValue', function() {
+      .stub(ConfigAggregator.prototype, 'getPropertyValue', () => {
         return TEST_CUSTOM_TEMPLATES_REPO;
       })
       .command(['force:apex:class:create', '--classname', 'foo'])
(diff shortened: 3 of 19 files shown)


packages/plugin-templates/test/utils/templateCommand.test.ts
```diff
@@ -120,12 +120,10 @@ describe('TemplateCommand', () => {
...
     test
       .withOrg()
-      .withProject()
+      //.withProject()
       .stdout()
-      .stub(ConfigAggregator.prototype, 'getPropertyValue', (key: string) => {
-        if (key === Config.CUSTOM_ORG_METADATA_TEMPLATES) {
-          return TEST_CUSTOM_TEMPLATES_REPO;
-        }
+      .stub(ConfigAggregator.prototype, 'getPropertyValue', function() {
```
gbockus-sf, line 125: Did you move from arrow functions to functions for any particular reason?

packages/plugin-templates/test/commands/force/analytics/template/create.test.ts
```diff
@@ -13,7 +13,7 @@ describe('Analytics template creation tests:', () => {
   describe('Check analytics template creation', () => {
     test
       .withOrg()
-      .withProject()
+      //.withProject() // temporarily broken in current version of command
```
gbockus-sf, line 16: Will we be able to release in this state or will that method being broken impact end users? If we need to fast follow to upgrade again when fixed should we create a ticket for it?

packages/plugin-templates/package.json
```diff
@@ -8,7 +8,7 @@
     "@oclif/command": "^1",
     "@oclif/config": "^1",
     "@oclif/errors": "^1",
-    "@salesforce/command": "^2.2.0",
+    "@salesforce/command": "^4.2.1",
```
gbockus-sf, line 11: Obviously at least 1 breaking change between 2 -> 4 ( withProject) Anything else we should be concerned about?

packages/plugin-templates/test/commands/force/analytics/template/create.test.ts
shetzel: I believe `.withProject` was broken in `@salesforce/command` (it was being stubbed twice), then fixed, then some recent oclif changes broke this (and more) again.  It's a TypeScript error that should be ok but of course not ideal.

shetzel (approved):

packages/plugin-templates/package.json
randi274: Tests are all running successfully now - next steps would be to spot check it with linking the plugin locally, and I can do that next.

packages/plugin-templates/test/utils/templateCommand.test.ts
randi274: Yes, but it had more to do with removing the parameters (which is really weird because getPropertyValue in [sfdx-core](https://github.com/forcedotcom/sfdx-core/blob/b77cdad56b114290b8e5fd23d968e23e7d4eedf4/src/config/configAggregator.ts#L161) definitely has a string parameter. I saw they were using function() instead of arrow functions in some CLI libraries, and was thinking that was the issue, but reviewing it now I think it was just an issue with the params.
Screen Shot 2022-01-25 at 2 22 31 PM
Screen Shot 2022-01-25 at 2 22 54 PM

packages/plugin-templates/package.json
gbockus-sf: sorry for the dumb q, but are there any release notes or anything for @salesforce/command? I found the current repo and it looks like they bumped to 4.0 when them moved repos, but I didn't see much else about breaking changes or anything.  Totally fine to gimme the 👋  if not an issue.

randi274: Not dumb at all, I don't know a ton about that repository! Here's the changelog, and @shetzel might be able to 👋  it off. https://github.com/salesforcecli/command/blob/main/CHANGELOG.md

gbockus-sf: thanks looks like the only breaking change called out is about moving some output from stderr to stdout so if all our tests are still passing probably not an issue for us.

packages/plugin-templates/test/commands/force/analytics/template/create.test.ts
randi274: FWIW the tests still pass as is - it would be great for us to verify withProject as well as an extra catch, but we can put a ticket in the backlog to revert this change once that's fixed. @shetzel do you have any concern for us removing withProject temporarily?

packages/plugin-templates/test/commands/force/analytics/template/create.test.ts
```diff
@@ -64,7 +64,7 @@ describe('Analytics template creation tests:', () => {
       );
     test
       .withOrg()
-      .withProject()
+      //.withProject()
```
gbockus-sf, line 67: Randi and I discussed commenting out vs deleting all these calls.  We'll have a dead code discussion at the team level.  Decided to leave commented out for now.

packages/plugin-templates/test/utils/templateCommand.test.ts
```diff
@@ -120,12 +120,10 @@ describe('TemplateCommand', () => {
...
     test
       .withOrg()
-      .withProject()
+      //.withProject()
       .stdout()
-      .stub(ConfigAggregator.prototype, 'getPropertyValue', (key: string) => {
-        if (key === Config.CUSTOM_ORG_METADATA_TEMPLATES) {
-          return TEST_CUSTOM_TEMPLATES_REPO;
-        }
+      .stub(ConfigAggregator.prototype, 'getPropertyValue', () => {
```
gbockus-sf, line 125: thanks

gbockus-sf (approved):
code changes look good 👍

PR #443, opened 2022-01-27
perf: move core to a careted version

randi274:
Allow our core dependency to keep up with small changes as needed


randi274:
✅  Spot checks looking good from linking plugin and calling create command in a few spots.

jag-sfdc (approved):

gbockus-sf (approved):
cool.  Can you update the description about why we're making this change?

PR #456, opened 2022-02-16
chore: use context for env variables

RitamAgrawal:
### What does this PR do?
Use context to define all environment variables

@W-10539922@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -258,7 +258,7 @@ workflows:
                 - main
       - run-win-tests
       - prepublish:
-          context: pdt-publish-restricted-context
+          context: IEAT_keys
           requires:
             - run-win-tests
             - node-latest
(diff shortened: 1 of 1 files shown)


mohanraj-r:
@RitamAgrawal Can you please try if this runs with [CircleCI Local CLI](https://circleci.com/docs/2.0/local-cli/)

.circleci/config.yml
```diff
@@ -273,15 +273,17 @@ workflows:
...
             branches:
               only: main
           requires:
             - prepublish
             - hold
       - publish-plugin:
-          context: pdt-publish-restricted-context
+          context:
+            - IEAT_keys
+            - AWS
```
gbockus-sf, line 286: So why is this one different?

mohanraj-r (approved):

.circleci/config.yml
RitamAgrawal: @gbockus-sf the only job that requires AWS credentials is publish-plugin. AWS credentials are defined in a separate context named AWS, which is shared by teams across forcedotcom.

gbockus-sf (commented):
👍

PR #353, opened 2021-07-01
Simplified linting scripts and reduced dependencies

pozil:
Following feedback from @pmdartus, this PR on the project template
- Merges the two linting scripts (Aura and LWC) into a single one
- Gets rid of `eslint-config-prettier` because it is not needed

PR #460, opened 2022-03-03
build: bumped deps for new project template

pozil:
### What does this PR do?

Updated node dependencies for new project template.


gbockus-sf (approved):
Passed circle builds. Thanks 👍

packages/templates/src/templates/project/package.json
```diff
@@ -16,18 +16,18 @@
     "precommit": "lint-staged"
   },
   "devDependencies": {
-    "@lwc/eslint-plugin-lwc": "^1.0.1",
-    "@prettier/plugin-xml": "^0.13.1",
```
mohanraj-r, line 20: @jag-sfdc I haven't worked with this repo much. Just wanted to do a quick sanity check .. 

- In terms of the changes in this PR, I am guessing the package.json being updated will be part of the template for new projects? 

- Would there be any concerns in bumping the dev deps across major versions - in terms of 
  	- compatibility with existing projects (guessing only new projects will get to use the updated package.json) or 
  	- mismatch with versions of the same deps used by other templates?

packages/templates/src/templates/project/package.json
```diff
@@ -16,18 +16,18 @@
...
-    "prettier": "^2.3.2",
+    "@salesforce/eslint-plugin-lightning": "^1.0.0",
+    "@salesforce/sfdx-lwc-jest": "^0.12.6",
+    "eslint": "^8.10.0",
+    "eslint-plugin-import": "^2.25.4",
+    "eslint-plugin-jest": "^26.1.1",
+    "husky": "^7.0.4",
+    "lint-staged": "^12.3.4",
+    "prettier": "^2.5.1",
     "prettier-plugin-apex": "^1.10.0"
```
mohanraj-r, line 31: @pozil Thanks for the PR. Is this PR updating to latest versions of all dev deps? Are there any deps that are not being updated to the latest available version?

pozil: @mohanraj-r yes, these are the latest available version.

packages/templates/src/templates/project/package.json
pozil: Yes to number one. I've already contributed to this project (PR #353 for example).

For number two, there shouldn't be incompatibilities with other projects since this is a standalone node project. The only risk would be to have incompatible dependencies within the project but it's not the case since we're already using those deps in our projects like [LWC Recipes](https://github.com/trailheadapps/lwc-recipes/blob/main/package.json).

packages/templates/src/templates/project/package.json
mohanraj-r: Thanks @pozil 
We will merge this PR after some [WIP templates changes](https://github.com/forcedotcom/salesforcedx-vscode/pull/3749) gets merged so we can test it together

mohanraj-r:
Superseded by #468

PR #464, opened 2022-03-11
build: remove unused yeoman environment dependency from create utils

mohanraj-r:
### What does this PR do?

- remove unused yeoman environment dependency from create utils
-  fix broken link to commands doc

@W-10832373@

### DIFF
--- a/packages/templates/src/utils/createUtil.ts
+++ b/packages/templates/src/utils/createUtil.ts
@@ -7,7 +7,6 @@
 import * as fs from 'fs';
 import * as path from 'path';
 // @ts-ignore
-import * as yeoman from 'yeoman-environment';
 import { nls } from '../i18n';
 
 /* tslint:disable:no-unused-expression */
--- a/packages/templates/src/utils/createUtil.ts
+++ b/packages/templates/src/utils/createUtil.ts
@@ -6,7 +6,6 @@
  */
 import * as fs from 'fs';
 import * as path from 'path';
-// @ts-ignore
 import { nls } from '../i18n';
 
 /* tslint:disable:no-unused-expression */
--- a/packages/templates/yarn.lock
+++ b/packages/templates/yarn.lock
@@ -4252,7 +4252,7 @@ shelljs@^0.7.3:
     interpret "^1.0.0"
     rechoir "^0.6.2"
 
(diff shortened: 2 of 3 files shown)


packages/templates/src/utils/createUtil.ts
```diff
@@ -7,7 +7,6 @@
 import * as fs from 'fs';
 import * as path from 'path';
 // @ts-ignore
```
gbockus-sf, line 9: nit: can you delete this ts-ignore as well?

packages/templates/yarn.lock
```diff
@@ -5151,7 +5151,7 @@ yeoman-environment@^2.9.5:
     untildify "^3.0.3"
     yeoman-generator "^4.8.2"
```
gbockus-sf, line 5152: this is somewhat confusing to me.  How does it still have `yeoman-generator "^4.8.2"` as a dep here but can be deleted below?

packages/templates/yarn.lock
```diff
@@ -5185,40 +5185,6 @@ yeoman-generator@4.11.0:
...
-    make-dir "^3.0.0"
-    mem-fs-editor "^7.0.1"
-    minimist "^1.2.5"
-    pretty-bytes "^5.2.0"
-    read-chunk "^3.2.0"
-    read-pkg-up "^5.0.0"
-    rimraf "^2.6.3"
-    run-async "^2.0.0"
-    semver "^7.2.1"
-    shelljs "^0.8.4"
```
gbockus-sf, line 5215: When I pulled down to to verify locally there was a change (to remove shelljs@^0.8.4) to the lock file file on a fresh install.  Can you commit that change?

packages/templates/yarn.lock
gbockus-sf: doesn't seem to beef one install 🤷

mohanraj-r: Think it is getting bumped to  4.11.0 following the semantic version range of `^`

gbockus-sf (changes requested):
Just a cleanup item.

packages/templates/yarn.lock
mohanraj-r: Done, I see it now too - I wasn't seeing it when I created the PR.

gbockus-sf (approved):
ty sir

PR #476, opened 2022-04-08
fix: update yeoman & shx to address high dependabot alerts.

gbockus-sf:
### What does this PR do?
Update Yeoman Generator and Yeoman environment to latest versions.  
Update shx.
Update code to account for breaking changes in yeoman and convert the logger to meet the requirements of the yeoman adapter.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,11 +5,7 @@
     "ecmaVersion": 6,
     "sourceType": "module"
   },
-  "plugins": [
-    "@typescript-eslint",
-    "jsdoc",
-    "eslint-plugin-header"
-  ],
+  "plugins": ["@typescript-eslint", "jsdoc", "eslint-plugin-header"],
   "extends": [
     "plugin:@typescript-eslint/recommended",
     "plugin:prettier/recommended"
--- a/package.json
+++ b/package.json
@@ -8,11 +8,10 @@
     "@commitlint/cli": "^16",
     "@commitlint/config-conventional": "^16",
     "@salesforce/dev-config": "1.4.1",
-    "commitizen": "^4.2.4",
-    "cz-conventional-changelog": "^3.3.0",
-    "typescript": "3.7.5",
     "@typescript-eslint/eslint-plugin": "^5.0.0",
     "@typescript-eslint/parser": "^2.22.0",
+    "commitizen": "^4.2.4",
+    "cz-conventional-changelog": "^3.3.0",
     "eslint": "^7.1.0",
     "eslint-config-prettier": "6.11.0",
     "eslint-plugin-header": "^3.0.0",
@@ -23,7 +22,8 @@
     "lint-staged": "^10.2.11",
     "mkdirp": "^0.5.1",
     "prettier": "1.19.1",
-    "pretty-quick": "^1.11.1"
+    "pretty-quick": "^1.11.1",
+    "typescript": "3.7.5"
   },
   "engines": {
     "node": ">=14.17.0"
--- a/packages/plugin-templates/package.json
+++ b/packages/plugin-templates/package.json
@@ -12,8 +12,8 @@
     "@salesforce/core": "^2.33.1",
     "@salesforce/templates": "54.3.0",
     "tslib": "^1",
-    "yeoman-environment": "2.4.0",
-    "yeoman-generator": "^4.8.2"
+    "yeoman-environment": "3.9.1",
+    "yeoman-generator": "^5.6.1"
   },
   "devDependencies": {
     "@oclif/dev-cli": "^1",
@@ -27,22 +27,22 @@
     "@types/mocha": "^5",
     "@types/node": "^10",
     "@types/yeoman-assert": "^3.1.1",
-    "@types/yeoman-generator": "4.11.0",
+    "@types/yeoman-generator": "^5.2.9",
     "chai": "^4",
     "eslint": "^8.4.1",
     "mocha": "^6.2.3",
     "nyc": "15.0.1",
-    "shx": "0.2.2",
+    "shx": "0.3.4",
     "sinon": "^7.3.1",
     "ts-node": "^8.6.2",
     "tslint": "^6.1.0",
     "typescript": "^3.8.3",
     "yeoman-assert": "^3.1.1",
-    "yeoman-test": "^2.0.0"
+    "yeoman-test": "^6.3.0"
   },
   "resolutions": {
-    "yeoman-generator": "4.11.0",
-    "@types/yeoman-generator": "4.11.0"
+    "yeoman-generator": "^5.6.1",
+    "@types/yeoman-generator": "^5.2.9"
   },
   "homepage": "https://github.com/forcedotcom/salesforcedx-templates",
   "keywords": [
--- a/packages/templates/src/generators/projectGenerator.ts
+++ b/packages/templates/src/generators/projectGenerator.ts
@@ -43,7 +43,10 @@ const analyticsVscodeExt = 'salesforce.analyticsdx-vscode';
 
 export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
   constructor(args: string | string[], options: ProjectOptions) {
-    super(args, options);
+    // Set customInstallTask to false so that the npm install will be attempted.
+    super(args, options, {
+      customInstallTask: false
+    });
     this.sourceRootWithPartialPath('project');
   }
 
--- a/packages/templates/src/generators/sfdxGenerator.ts
+++ b/packages/templates/src/generators/sfdxGenerator.ts
@@ -27,8 +27,22 @@ export abstract class SfdxGenerator<
    */
   public builtInTemplatesRootPath?: string;
 
-  constructor(args: string | string[], options: TOptions) {
-    super(args, options);
+  /**
+   * The constructor for the SFDXGenerator.
+   *
+   * @param args arguments to pass to the yeoman generator constructor.
+   * @param options SFDXGenerator specific options.
+   * @param features The yeoman GeneratorFeatures.  Defaults to a customInstallTask of false in order to skip the automatic npm/yarn install.  Override to true if you need to run the package manager install.
+   */
+  constructor(
+    args: string | string[],
+    options: TOptions,
+    features: Generator.GeneratorFeatures = {
+      customInstallTask: true
+    }
+  ) {
+    // custom install task set to true keeps us from getting a warning message about a missing package.json file.
+    super(args, options, features);
     this.options.apiversion =
       this.options.apiversion ?? TemplateService.getDefaultApiVersion();
     this.options.outputdir = this.options.outputdir ?? process.cwd();
--- a/packages/templates/src/generators/sfdxGenerator.ts
+++ b/packages/templates/src/generators/sfdxGenerator.ts
@@ -32,22 +32,24 @@ export abstract class SfdxGenerator<
    *
    * @param args arguments to pass to the yeoman generator constructor.
    * @param options SFDXGenerator specific options.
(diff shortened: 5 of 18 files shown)


.eslintrc.json
```diff
@@ -5,11 +5,7 @@
     "ecmaVersion": 6,
     "sourceType": "module"
   },
-  "plugins": [
-    "@typescript-eslint",
-    "jsdoc",
-    "eslint-plugin-header"
-  ],
+  "plugins": ["@typescript-eslint", "jsdoc", "eslint-plugin-header"],
```
gbockus-sf, line 8: There was a reformat script I ran at the top level that reformatted a few files.

package.json
```diff
@@ -23,7 +22,8 @@
     "lint-staged": "^10.2.11",
     "mkdirp": "^0.5.1",
     "prettier": "1.19.1",
-    "pretty-quick": "^1.11.1"
+    "pretty-quick": "^1.11.1",
+    "typescript": "3.7.5"
```
gbockus-sf, line 26: running yarn alphabetized these.

packages/plugin-templates/package.json
```diff
@@ -12,8 +12,8 @@
     "@salesforce/core": "^2.33.1",
     "@salesforce/templates": "54.3.0",
     "tslib": "^1",
-    "yeoman-environment": "2.4.0",
-    "yeoman-generator": "^4.8.2"
+    "yeoman-environment": "3.9.1",
+    "yeoman-generator": "^5.6.1"
```
gbockus-sf, line 16: big change here was bumping these the latest release of each.

packages/plugin-templates/package.json
```diff
@@ -27,22 +27,22 @@
...
     "@types/node": "^10",
     "@types/yeoman-assert": "^3.1.1",
-    "@types/yeoman-generator": "4.11.0",
+    "@types/yeoman-generator": "^5.2.9",
     "chai": "^4",
     "eslint": "^8.4.1",
     "mocha": "^6.2.3",
     "nyc": "15.0.1",
-    "shx": "0.2.2",
+    "shx": "0.3.4",
```
gbockus-sf, line 35: address another high dependabot alert.

packages/templates/src/generators/projectGenerator.ts
```diff
@@ -43,7 +43,10 @@ const analyticsVscodeExt = 'salesforce.analyticsdx-vscode';
 
 export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
   constructor(args: string | string[], options: ProjectOptions) {
-    super(args, options);
+    // Set customInstallTask to false so that the npm install will be attempted.
+    super(args, options, {
+      customInstallTask: false
+    });
```
gbockus-sf, line 49: This was pretty tricky.  Basically the yeoman flow has changed how it checks if an npm install is needed.  For most of our templates we don't have a package.json so it was printing an error message that I figured we would want to avoid showing to our end users.  
The message was `No change to package.json was detected. No package manager install will be executed`

I didn't find any yeoman documentation around this just had to walk the source to find the work around.  If we set customInstallTask to true it will early exit from the npm install check w/o printing an error message.

packages/templates/src/generators/sfdxGenerator.ts
```diff
@@ -27,13 +27,29 @@ export abstract class SfdxGenerator<
...
+   * @param features The yeoman GeneratorFeatures. Defaults to a customInstallTask of false in order to skip the automatic npm/yarn install. Override to true if you need to run the package manager install.
+   */
+  constructor(
+    args: string | string[],
+    options: TOptions,
+    // custom install task set to true keeps us from getting a warning message about a missing package.json file.
+    // Note for generators that need an npm install it should be set to false.
+    features: Generator.GeneratorFeatures = {
+      customInstallTask: true
+    }
```
gbockus-sf, line 44: The one template we have that needs the npm install is the project template.  Hence I'm overriding the default value here.

packages/templates/src/service/templateService.ts
```diff
@@ -109,23 +109,25 @@ export class TemplateService {
     this.adapter.log.clear();
 
     return new Promise((resolve, reject) => {
-      this.env.run(generatorNamespace, templateOptions, err => {
-        if (err) {
```
gbockus-sf, line 113: The move to the promise based run call was the breaking change in yeoman between version 4 -> 5.

packages/templates/src/utils/adapter.ts
```diff
@@ -5,11 +5,12 @@
...
  */
 import * as generator from 'yeoman-generator';
-import { Log } from './logger';
+import { getYeomanLogger, Log } from './logger';
 import { Answers } from './types';
 
 export class ForceGeneratorAdapter {
-  public log = new Log();
+  private _log = new Log();
+  public log = getYeomanLogger(this._log);
```
gbockus-sf, line 13: This was super fun.  Yeoman expect the adapter log to be implemented a specific way.  Previously we didn't follow their example so the log calls were failing.  I have a comment in logger.ts calling out where to read about it in the yeoman source.

packages/templates/src/utils/logger.ts
```diff
@@ -72,3 +72,102 @@ export class Log {
...
+  getCleanOutput(): string[];
+  getOutput(): string;
+  skip(arg: string): ILogRef;
+  force(arg: string): ILogRef;
+  create(arg: string): ILogRef;
+  invoke(arg: string): ILogRef;
+  conflict(arg: string): ILogRef;
+  identical(arg: string): ILogRef;
+  info(arg: string): ILogRef;
+}
```
gbockus-sf, line 92: There is probably a cleve way to do this in TS but I went with the plain but obvious interface for the logger we return.

packages/templates/src/utils/logger.ts
```diff
@@ -72,3 +72,102 @@ export class Log {
     this.cleanOutput = [];
   }
 }
+
+export interface ILogRef {
+  (args: string): void;
```
gbockus-sf, line 77: You read that right...it's both a function and an object

packages/templates/src/utils/logger.ts
```diff
@@ -72,3 +72,102 @@ export class Log {
...
+
+  yeomanLogger.identical = (args: string) => {
+    log.identical(args);
+    return yeomanLogger;
+  };
+
+  yeomanLogger.info = (args: string) => {
+    log.info(args);
+    return yeomanLogger;
+  };
```
gbockus-sf, line 141: In the Log class implementation we iterate over a list of sting to add as properties.  I tried creating a representation of this in the interface, but was never successful.

packages/templates/test/generators/sfdxGenerator.test.ts
```diff
@@ -23,10 +24,10 @@ describe('SfdxGenerator', () => {
     }
     public doWriting(options: object) {}
   }
+  const testEnv = YeomanEnvironment.createEnv();
+  testEnv.cwd = process.cwd();
   const mockMyGeneratorOptions = {
-    env: {
-      cwd: process.cwd()
-    },
+    env: testEnv,
```
gbockus-sf, line 30: This test started failing due to missing properties in the env.  I went the easy route over the any route.

packages/templates/test/service/templateService.test.ts
```diff
@@ -397,36 +397,41 @@ describe('TemplateService', () => {
...
+      )}\n   create ${path.normalize(
+        'testsoutput/libraryCreate/apexClass/LibraryCreateClass.cls'
+      )}\n   create ${path.normalize(
+        'testsoutput/libraryCreate/apexClass/LibraryCreateClass.cls-meta.xml'
+      )}\n`;
+
+      expect(result.rawOutput).to.equal(
+        actual,
+        'Actual property did not match'
+      );
```
gbockus-sf, line 430: I broke these into separate expects in order to make it easy to track down what had changed.  Originally when just a single property was failing it was difficult to narrow down exactly what didn't match.

packages/templates/test/utils/logger.test.ts
```diff
@@ -48,3 +50,127 @@ describe('write', () => {
...
+    expect(yeomanLogger).not.to.equal(undefined);
+    yeomanLogger(message);
+    assert.calledOnce(infoStub);
+    assert.calledWith(infoStub, message);
+  });
+
+  it('info()', () => {
+    const result = yeomanLogger.info(message);
+    expect(result).to.eql(yeomanLogger);
+    assert.calledWith(infoStub, message);
```
gbockus-sf, line 78: I sort of hated mixing the assert and expect, but this mocha install was interesting. `expect(stub).to.be.calledWith` wasn't recognized as a function.

packages/plugin-templates/package.json
mohanraj-r: any reasoning behind keeping the yeoman dependencies in mixed pinned states ?

packages/templates/src/generators/projectGenerator.ts
mohanraj-r: Great find 👍 

- Can we move some more of this PR comment as code comment e.g. "For most of our templates we don't have a package.json, hence .. " 
- Should "the npm install will be attempted" be "the npm install will not be attempted"? Guess I am a little confused.

packages/templates/test/service/templateService.test.ts
```diff
@@ -397,36 +397,41 @@ describe('TemplateService', () => {
...
-          process.cwd(),
-          'testsoutput/libraryCreate/apexClass'
-        ),
-        created: [
+
+      expect(result.outputDir).to.equal(
+        path.resolve(process.cwd(), 'testsoutput/libraryCreate/apexClass'),
+        'outputDir property did not match'
+      );
+
```
mohanraj-r, line 405: Would it make sense to add a test exercising the custom install flag ?

mohanraj-r (commented):
Looks good. Some minor comments/questions.

mohanraj-r:
Can you also mark/close the open PRs by dependabot, snyk that is taken care of in this PR.

packages/templates/src/service/templateService.ts
smithgp: Sorry to jump in and this is a bit of a nit, but, since `run()` now returns a Promise, you don't really need the extra outer Promise and resolve/reject callbacks, and could just do it as an async/await, e.g.:
```
await this.env.run(generatorNamespace, templateOptions);
// the body of the then() here...
return result;
// and you don't need a try/catch since you were doing anything special with any errors
```

mohanraj-r (approved):

mohanraj-r:
Due to the time urgency of getting the sec fixes in this PR released and there being no blocking/major comments that need to be addressed (and @gbockus-sf being out today) merging this PR as is. FYI @smithgp

packages/plugin-templates/package.json
gbockus-sf: nope. Was just keeping what was in place there.  Have we landed on pinned vs not?  Seems like the latest pain led us back to pinning right?

packages/templates/src/generators/projectGenerator.ts
gbockus-sf: question 1: Yup will update in a new PR.
on question 2: Yeah, so the default value [here](https://github.com/forcedotcom/salesforcedx-templates/blob/5e358c53086e9af61aa2b8613a3aea7c05a6d730/packages/templates/src/generators/sfdxGenerator.ts#L43) is setting the property to true which results in skipping the npm install. For this class we set it to false b/c it needs has a package.json and needs the npm install.

packages/templates/src/service/templateService.ts
gbockus-sf: @smithgp jump in any time.  Will update in a follow on PR.  Good call and appreciate the feedback.

PR #486, opened 2022-05-06
refactor: address feedback from yeoman update PR #476

gbockus-sf:
### What does this PR do?
- Address feedback in PR for upgrading Yeoman
  - pin yeoman version
  - Move env.run call to use await/async
  - Add tests for projectGenerator and customInstallTask parameter.

### DIFF
--- a/packages/templates/src/service/templateService.ts
+++ b/packages/templates/src/service/templateService.ts
@@ -108,27 +108,19 @@ export class TemplateService {
 
     this.adapter.log.clear();
 
-    return new Promise((resolve, reject) => {
-      this.env
-        .run(generatorNamespace, templateOptions)
-        .then(() => {
-          const outputDir = path.resolve(this.cwd, templateOptions.outputdir!);
-          const created = this.adapter.log.getCleanOutput();
-          const rawOutput = nls.localize('RawOutput', [
-            outputDir,
-            this.adapter.log.getOutput()
-          ]);
-          const result = {
-            outputDir,
-            created,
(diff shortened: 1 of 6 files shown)


packages/templates/src/service/templateService.ts
```diff
@@ -108,27 +108,19 @@ export class TemplateService {
...
-            created,
-            rawOutput
-          };
-          resolve(result);
-        })
-        .catch(err => {
-          reject(err);
-        });
-    });
+    await this.env.run(generatorNamespace, templateOptions);
```
gbockus-sf, line 111: same functionality with await/async

packages/templates/test/generators/projectGenerator.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    checkInputsStub = stub(CreateUtil, 'checkInputs');
+    mkdirSyncStub = stub(fs, 'mkdirSync').returns(undefined);
+    existsSyncStub = stub(fs, 'existsSync').returns(false);
+  });
+
+  afterEach(() => {
+    restore();
+  });
+
+  it('Should set the customInstallTask feature to false for the ProjectGenerator', () => {
```
gbockus-sf, line 42: went to add just this test, but then much us the file was uncovered so I went ahead and added some simple tests to cover the rest of the functionality.

smithgp (approved):

mshanemc:
we dropped yeoman

PR #491, opened 2022-05-26
Added unit test to the LWC template

jeffb-sfdc:
### What does this PR do?
The PR adds a default/boilerplate unit test to the LWC template.

When a new LWC is created, only the .js and .html are created. We should encourage our customers to write unit tests, and adding a default unit test along with the component class will help encourage customers to write unit tests.

What issues does this PR fix or reference?
#@W-11171762@


packages/templates/src/generators/lightningComponentGenerator.ts
```diff
@@ -131,40 +131,77 @@ export default class LightningComponentGenerator extends SfdxGenerator<
...
-        this.destinationPath(path.join(outputdir, fileName, `${fileName}.js`)),
-        { className }
+        this.destinationPath(
+          path.join(
+            outputdir,
+            camelCaseComponentName,
+            `${camelCaseComponentName}.js`
+          )
+        ),
+        { pascalCaseComponentName }
```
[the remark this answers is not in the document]
jeffb-sfdc: Thanks @smithgp !  I re-ran the unit tests locally and found a few errors (like this).  This has been fixed.

packages/templates/src/generators/lightningComponentGenerator.ts
```diff
@@ -131,40 +131,77 @@ export default class LightningComponentGenerator extends SfdxGenerator<
...
+            outputdir,
+            camelCaseComponentName,
+            `${camelCaseComponentName}.html`
+          )
         ),
         {}
       );
+
+      this.fs.copyTpl(
+        this.templatePath(path.join(`__tests__`, `${template}.test.js`)),
```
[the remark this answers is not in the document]
jeffb-sfdc: They have been added.  If you'd like any changes in any of the templates, please let me know.

smithgp (approved):
Excellent, thanks for adding this, @jeffb-sfdc

packages/templates/src/templates/lightningcomponent/lwc/analyticsDashboardWithStep/__tests__/analyticsDashboardWithStep.test.js
```diff
@@ -0,0 +1,25 @@
...
+        const element = createElement('c-<%= kebabCaseComponentName %>', {
+            is: <%= pascalCaseComponentName %>
+        });
+
+        // Act
+        document.body.appendChild(element);
+
+        // Assert
+        // const div = element.shadowRoot.querySelector('div');
+        expect(1).toBe(2);
```
jeffb-sfdc, line 23: I'll change this to `expect(1).to.equal(1);` so it passes out of the box.

smithgp: Is `to.equal()` right? Isn't that `chai` style expect? I though in jest it's `.toEqual()`.

jeffb-sfdc: Thanks @smithgp !  I've updated them to `expect(1).toEqual(1);`

randi274 (commented):
This is looking good! I was able to build locally and verified that:
✅ Creating a new component also created a new test directory + file successfully
✅ Running that new test passed without having to make other changes

I did notice was that the test file we're creating when we run the `Create Lightning Web Component Test` is different than this new file that's created in the component create command. Is it possible to have them be the same?
Screen Shot 2022-06-01 at 11 45 36 AM

Additionally, is there any way to combine the logic for the LightningTestGenerator with this new code? Do we want or need to? Just a thought I had as I was reviewing the flow for creating the test file separately.

jeffb-sfdc:
@randi274 I changed the three templates to `expect(1).toBe(1)`.  As far as consolidation, I'd prefer to keep separate.

packages/templates/src/templates/lightningcomponent/lwc/analyticsDashboardWithStep/__tests__/analyticsDashboardWithStep.test.js
jeffb-sfdc: @smithgp based on feedback from @randi274 , I changed them each to use `toBe()`.  Though I doubt there are, if there are any issues using toBe() over toEqual(), please let me know.

PR #509, opened 2022-09-09
chore: remove listing git tags so publish will not fail

gbockus-sf:
### What does this PR do?
Removes the unnecessary listing of the git tags so that auto publish jobs will complete. 

Found when publish was attempted: https://app.circleci.com/pipelines/github/forcedotcom/salesforcedx-templates/1589/workflows/1600396c-c065-4125-8c7a-28cf09f1ad31/jobs/6155

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,9 +83,6 @@ commands:
     steps:
       - gh-config
       - run:
-          name: Current git tags
-          command: git tag
-      - run:
           name: Create new git tag
           command: |
             PKG_TAG="v$(cat package.json | jq -r .version)"


.circleci/config.yml
```diff
@@ -82,9 +82,6 @@ commands:
   create-git-tag:
     steps:
       - gh-config
-      - run:
-          name: Current git tags
-          command: git tag
```
gbockus-sf, line 87: This is just listing the current git tags.  It was running out of room and depending on user feedback to finish printing all the tags so it always failed in circleci.

klewis-sfdc (approved):
LGTM!

PR #516, opened 2022-10-12
chore: switch CI to Github Actions

randi274:
### What does this PR do?
Removes CircleCI configuration, documentation, and tasks in favor of Github Actions.

New Github Workflows:
- `test.yml`: Runs unit tests on every commit, except to main.
- `onPushToMain.yml` - On push to main, bumps the version and creates a tag and release in github.
- `onRelease.yml` - On successful creation of the release, publishes the library and the plugin to NPM.
- `failureNotification.yml` - if the previous two workflows fail, alert us in Slack using the IDEE webhook.
- `validatePR.yml` - ensure PRs have WIs associated.

@W-11596496@

### Notes
The first pass at this automation was created by the valet runner with #515. This was a good start, but it would have been all to easy for us to fall into the same pattern of building overly long singular jobs. Instead, fans of the Salesforce CLI may notice  that this change is shockingly similar to the new automation set forth in the plugin-analytics [repo](https://github.com/salesforcecli/plugin-analytics/pull/138). This is very much by design. Not only do we wish to keep our library in lockstep with the CLI, but the patterns established in that PR demonstrate a better understanding of the key attributes of github actions that we should also seek to leverage. The workflows are small, tight, and delegate a majority of the work to a shared space in the salesforcecli repository. If we find ourselves unable or unwilling to call those workflows directly, we would do well to create a shared forcedotcom actions library ourselves and continue to delegate the work separately.


.github/actions/setup-publish/action.yml
```diff
@@ -0,0 +1,36 @@
...
+runs:
+  using: composite
+  steps:
+  - uses: "./.github/actions/gh-config"
+  - uses: actions/checkout@v2
+  - name: Setup Node.js
+    uses: actions/setup-node@v3
+    with:
+      node-version: latest
+      cache: 'yarn'
```
randi274, line 13: I've seen different answers on whether this actually caches node modules or not.

.github/workflows/onRelease.yml
```diff
@@ -0,0 +1,78 @@
...
+      run: |-
+        RELEASE_VERSION="$(node -pe "require('./package.json').version")"
+        TARFILE=$(ls | grep $RELEASE_VERSION.tgz)
+        echo "publishing to npm using ${TARFILE}"
+        npm publish $TARFILE --access public
+
+  publish-plugin:
+    defaults:
+      run:
+        working-directory: "~/cli/packages/plugin-templates"
```
randi274, line 64: I'm *hopeful* that this is all we have to do to setup for the `npmPublish` to work correctly - would really love review by the CLI team to take a look and see if this gut instinct is right or not.

.github/workflows/test.yml
```diff
@@ -0,0 +1,10 @@
+name: test
+
+on:
+  push:
+    branches-ignore: [main]
+  workflow_dispatch:
+
+jobs:
+  unit-tests:
+    uses: salesforcecli/github-workflows/.github/workflows/unitTest.yml@main
```
randi274, line 10: If there's any reason we shouldn't be calling the CLI workflow for this, we could easily add the build + test commands here, but I couldn't think of any, and this is way simpler.

.github/actions/setup-publish/action.yml
```diff
@@ -0,0 +1,36 @@
...
+  - name: Install dependencies
+    run: |-
+      yarn --version
```
mshanemc, line 16: gha already provides that info in the logs from `setup-node`
https://github.com/salesforcecli/testPackageRelease/actions/runs/3237428684/jobs/5304483090#step:3:10
https://github.com/salesforcecli/testPackageRelease/actions/runs/3237428684/jobs/5304483090#step:3:14

.github/workflows/onRelease.yml
mshanemc: I think it'll run fine from the subfolder.

.github/workflows/onRelease.yml
```diff
@@ -0,0 +1,78 @@
...
+      image: node:16
+    needs:
+    - publish-library
+    steps:
+    - name: CLI NPM Publish
+      uses: salesforcecli/github-workflows/.github/workflows/npmPublish.yml@main
+      with:
+        ctc: true
+        sign: true
+        tag: latest
```
mshanemc, line 76: `tag: latest` is optional (it's the default)

.github/workflows/onRelease.yml
```diff
@@ -0,0 +1,78 @@
...
+        githubTag: ${{ github.event.release.tag_name || inputs.tag }}
```
mshanemc, line 77: one reason this might not work: the publish workflow is gonna check out the githubTag.

That may not have your version bump from publishing the library (ugh, lerna).  Even if you commit it, it'll won't be on that tag.

mshanemc: It doesn't look like y'all have used Releases before: https://github.com/forcedotcom/salesforcedx-templates/releases

So I can't really see what you've been doing.

It's possible that lerna bumps the versions in pjson and then sends the release, but I'm not sure.

.github/workflows/onPushToMain.yml
```diff
@@ -0,0 +1,10 @@
+name: version, tag and github release
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  release:
+    uses: salesforcecli/github-workflows/.github/workflows/githubRelease.yml@main
```
mshanemc, line 9: this action is designed to produce a github release based on past release changelogs, the commits, and the current pjson.

I don't know that it'll work for lerna since it would need to increment multiple package.json

.github/workflows/onRelease.yml
randi274: Ok cool - so it sounds like if we can get the github release added to our workflow, make sure the packages are all incrementing correctly there even w/ lerna, we should *hopefully* be good to invoke this action. 🤞

.github/workflows/onPushToMain.yml
randi274: Indeed, this does seem to perform incorrectly, so I'll see if I can re-write that part and revisit once that's looking better: https://github.com/randi274/salesforcedx-templates/commit/b8af9f8bc23ac959a560c91e7e6051e2978f30ff

contributing/publishing.md
```diff
@@ -17,13 +16,13 @@ The salesforcedx-templates project uses a two-branch strategy. Work that is curr
 
 ## Publishing to NPM
 
-To publish the changes to npm, we run the task `Publish Library and Plugin`. This task calls the script `publish-workflow.sh` and prompts the user for the required information. The publish-workflow script generates an HTTP Request to the CircleCI API. It tells CircleCI that it wants to run the `publish-workflow` from the `main` branch.
+When a commit is merged to main, we will automatically create the github release, and then publish the changes to npm using our Github Actions.
```
klewis-sfdc, line 19: 👍🏽

.github/workflows/onRelease.yml
```diff
@@ -0,0 +1,78 @@
...
+    - name: Set .npmrc
+      run: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
+    - name: Publish to NPM
+      run: |-
+        RELEASE_VERSION="$(node -pe "require('./package.json').version")"
+        TARFILE=$(ls | grep $RELEASE_VERSION.tgz)
+        echo "publishing to npm using ${TARFILE}"
+        npm publish $TARFILE --access public
+
+  publish-plugin:
```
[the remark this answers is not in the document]
klewis-sfdc: Yeah, for that reason it seems like breaking them up might be better?  So that you could easily re-run publish-plugin only in the case that it fails after publish library succeeds...

.github/workflows/onRelease.yml
```diff
@@ -0,0 +1,78 @@
...
+        required: true
+
+jobs:
+  prepublish:
+    defaults:
+      run:
+        working-directory: "~/cli"
+    runs-on: ubuntu-latest
+    container:
+      image: node:16
```
klewis-sfdc, line 21: Noticed that the old publish-library and publish-plugin circle flows used node:14.  Just confirming that the bump to 16 is intentional here?

.github/workflows/test.yml
klewis-sfdc: Are the workflows that are shared generic enough that we are confident that there won't be changes to them that would have a negative impact on this repo's usage of them?  Are the workflows that are being reused from the salesforcecli repo here intended to be re-used from outside of the salesforcecli github account?

.github/workflows/onRelease.yml
randi274: It was, but it probably shouldn't be done before updating the rest! I'll revert this for now.

.github/workflows/test.yml
randi274: This is a good question! I think they're safe to be run for any CLI plugin that's also making use of yarn as their package manager. You can see the list here - we can also pin at a specific commit hash, instead of using main if you think this is something we should guard against. https://github.com/salesforcecli/github-workflows/blob/main/.github/workflows/unitTest.yml

package.json
```diff
@@ -30,6 +30,7 @@
     "nyc": "^15.1.0",
     "prettier": "^2.0.5",
     "pretty-quick": "^3.1.0",
+    "shelljs": "^0.8.5",
```
randi274, line 33: This was already an indirect dependency for us; I added this directly so the precommit hook allows lerna to bump the necessary versions.

scripts/bump-lerna-version.js
```diff
@@ -0,0 +1,22 @@
...
+
+const shell = require('shelljs');
+
+export function preCommit(props) {
+  if (props && props.version) {
+    shell.exec(`yarn`);
+    shell.exec(
+      `yarn lerna version ${props.version} --force-publish --no-git-tag-version --exact --yes`
+    );
+    shell.exec(`yarn install --ignore-scripts --package-lock-only --no-audit`);
```
randi274, line 16: This doesn't seem to be having the desired effect of updating the package lock for us... we might have to consider another path.

.github/workflows/onPushToMain.yml
```diff
@@ -0,0 +1,40 @@
+name: version, tag and github release
+# todo: after this repo is split out and lerna is removed, return to this gha: 
+# https://github.com/salesforcecli/github-workflows/blob/main/.github/workflows/githubRelease.yml
```
randi274, line 3: @mshanemc - I re-wrote the onPushToMain to make sure that lerna version bump is happening correctly. Does this seem like it should be fine downstream to work with npmPublish of the plugin now?

Here's a successful run of that in action as well: 
- Workflow: https://github.com/randi274/salesforcedx-templates/actions/runs/3252787652/workflow 
- Commit: https://github.com/randi274/salesforcedx-templates/commit/9227350132e881f941a98aa25001a1e087fed8a8
- Corresponding tag/release (which is pretty noisy since i didn't have a release there yet): https://github.com/randi274/salesforcedx-templates/releases/tag/55.2.0

.github/workflows/onRelease.yml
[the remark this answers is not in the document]
randi274: Done!

.github/workflows/onPushToMain.yml
randi274: Although I'm getting a bad feeling that this won't "just work" since I'm currently having it publish both packages... https://github.com/salesforcecli/github-workflows/blob/main/.github/workflows/npmPublish.yml#L98

packages/plugin-templates/README.md
```diff
@@ -2,7 +2,6 @@
 
 This repository provides a series of commands, templates, and generators for various metadata types. It is bundled with the [salesforcedx plugin](https://www.npmjs.com/package/salesforcedx).
```
klewis-sfdc, line 4: Does the "salesforce plugin" link above direct to the correct place?  Seems like maybe it should point here instead? https://github.com/salesforcecli/status.  Asking because I see this same link in the apex repo.

.github/actions/setup-publish/action.yml
```diff
@@ -0,0 +1,25 @@
...
+    shell: bash
+  - name: Build project
+    run: yarn build
+    shell: bash
+  - name: Generate plugin manifest
+    run: yarn manifest:generate
+    shell: bash
+  - name: Pack Library
+    run: npm pack
+    working-directory: "./packages/templates"
```
randi274, line 24: Need to update this for new structure

.github/workflows/failureNotification.yml
```diff
@@ -0,0 +1,40 @@
...
+  workflow_run:
+    workflows:
+      - version, tag and github release
+      - publish
+    types:
+      - completed
+
+jobs:
+  failure-notify:
+    runs-on: ubuntu-latest
```
randi274, line 13: Need to change this to match what we're doing in LLS

.github/workflows/onRelease.yml
```diff
@@ -0,0 +1,67 @@
...
+    runs-on: ubuntu-latest
+    container:
+      image: node:14
+    steps:
+    - uses: "./.github/actions/setup-publish"
+    - uses: actions/upload-artifact@v2
+      with:
+        path: |-
+          ./.git
+          ./packages/templates/package.json
```
randi274, line 28: Change for new structure

.github/workflows/onRelease.yml
```diff
@@ -0,0 +1,67 @@
...
+    - name: Set .npmrc
+      run: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
+    - name: Publish to NPM
+      run: |-
+        RELEASE_VERSION="$(node -pe "require('./package.json').version")"
+        TARFILE=$(ls | grep $RELEASE_VERSION.tgz)
+        echo "publishing to npm using ${TARFILE}"
+        npm publish $TARFILE --access public
+
+  publish-plugin:
```
randi274, line 61: this can all be removed now

.github/workflows/publishPlugin.yml
```diff
@@ -0,0 +1,33 @@
+name: publish plugin
```
randi274, line 1: Remove me

package.json
randi274: Actually, since we can probably just use standard versioning now, I don't think this is necessary either.

scripts/bump-lerna-version.js
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+const shell = require('shelljs');
```
randi274, line 8: I think I can remove this too. Might need some testing to confirm.

.github/actions/setup-publish/action.yml
```diff
@@ -0,0 +1,25 @@
...
+  - run: yarn
+    shell: bash
+  - name: Build project
+    run: yarn build
+    shell: bash
+  - name: Generate plugin manifest
+    run: yarn manifest:generate
+    shell: bash
+  - name: Pack Library
+    run: npm pack
```
randi274, line 23: and maybe yarn insead

randi274:
Merging main cleanly was proving problematic, so I'm closing this PR in favor of #527

PR #520, opened 2022-10-31
feat: separate plugin-templates from this repository

randi274:
### What does this PR do?
Removes the plugin-templates library from this repository. Plugin-templates will now be a stand-alone repository [here](https://github.com/salesforcecli/plugin-templates). 

@W-11309541@


package.json
```diff
@@ -41,6 +41,7 @@
     "chai": "^4.2.0",
     "chai-as-promised": "^7.1.1",
     "commitizen": "^4.2.4",
+    "cz-conventional-changelog": "^3.3.0",
```
klewis-sfdc, line 44: Is this something I should also add to the apex-node library?

package.json
```diff
@@ -21,53 +50,31 @@
...
     "prettier": "^2.0.5",
     "pretty-quick": "^3.1.0",
+    "shx": "^0.3.4",
     "sinon": "10.0.0",
     "ts-node": "^10.0.0",
-    "typescript": "^4.4.4"
+    "typescript": "^4.4.4",
+    "yeoman-assert": "^3.1.1"
   },
-  "engines": {
```
gbockus-sf, line 37: You should probably put engines back.

tsconfig.settings.json
```diff
@@ -11,6 +11,7 @@
     "outDir": "./lib",
     "preserveConstEnums": true,
     "strict": true,
+    "allowSyntheticDefaultImports": true,
```
gbockus-sf, line 14: did something change resulting in this addition?

.nycrc
```diff
@@ -1,16 +1,5 @@
...
-  ],
-  "cache": false,
-  "include": [
-    "lib**",
-    "src**"
-  ]
-}
\ No newline at end of file
+  "nyc": {
+    "extends": "@salesforce/dev-config/nyc"
```
gbockus-sf, line 3: does this do that ☝️

gbockus-sf (commented):
Ran though the PR as best I could and had a couple of trivial suggestions.  Pulled local and verified all the build pieces.  Only issue is lint isn't working currently 

```
question Which command would you like to run?: lint
$ eslint -c .eslintrc.json --ext .ts ./src ./test

Oops! Something went wrong! :(

ESLint: 7.32.0

ESLint couldn't find the config "../.eslintrc.js" to extend from. Please check that the name of the config is correct.

The config "../.eslintrc.js" was referenced from the config file in "/Users/gbockus/github/PDT/salesforcedx-templates/test/.eslintrc.js".

If you still have problems, please stop by https://eslint.org/chat/help to chat with the team.

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
ansible|gbockus-ltm at ~/github/PDT/salesforcedx-templates ±(randi/remove-plugin) ❯```

File: package.json [hunk 5]
```diff
@@ -21,53 +50,31 @@
...
-    "node": ">=14.17.0"
+  "resolutions": {
+    "yeoman-generator": "^5.6.1",
+    "@types/yeoman-generator": "^5.2.9"
```
gbockus-sf, line 68: this may no longer be necessary

tsconfig.settings.json
randi274: this was present in the original templates-level file that I had moved to the top level, so I think the diff view is a little confused: https://github.com/forcedotcom/salesforcedx-templates/blob/8cb2f51d25a054db61891c52f5ad24802677ce1c/packages/templates/tsconfig.json

.nycrc
randi274: this was also present at the templates-level initially! I think Willie had added it in a previous PR

package.json
randi274: I would be super nervous to remove yeoman in here... do you think it's not necessary because it was used in plugin-templates mostly before?

.nycrc
randi274: And to actually answer the question ... no. https://github.com/forcedotcom/dev-config/blob/main/nyc.json 

I'm thinking I might want to revert in that case?

randi274:
@gbockus-sf linting error was fixed - also I don't think linting was running automatically and I got a ton of errors when I ran the command, so I fixed those too. I'll look into seeing if we can get it to run on commits like we do with vs code.

package.json
gbockus-sf: This is just the [resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) block. It basically ties us & the deps to a version of yeoman.  It's fine we can leave it.

.nycrc
gbockus-sf: yeah agreed, probably want to go with what we had originally

package.json
```diff
@@ -21,53 +51,38 @@
...
+    "compile": "tsc -b",
+    "lint": "eslint -c .eslintrc.json --ext .ts ./src ./test",
+    "lint:fix": "eslint -c .eslintrc.json --ext .ts ./src ./test --fix",
+    "test": "nyc --exclude src/i18n/localization.ts --extension .ts mocha --parallel --forbid-only \"test/**/*.test.ts\""
   },
-  "lint-staged": {
-    "./{src,test}/**/*.{ts,js}": [
-      "eslint -c .eslintrc.json --fix"
-    ]
+  "husky": {
```
randi274, line 81: ok so adding this was pretty cool. We now get the following when we commit, and then a full lint + test happened on push! 
Screen Shot 2022-11-03 at 4 59 22 PM

.nycrc
```diff
@@ -4,13 +4,7 @@
   "statements": 75,
   "functions": 75,
   "branches": 75,
-  "reporter": [
-    "text",
-    "lcov"
-  ],
+  "reporter": ["text", "lcov"],
```
randi274, line 7: This is part of the commit hook now, so it stays this way

package.json
gbockus-sf: 🏆

gbockus-sf (approved):
🚢

PR #527, opened 2022-11-15
chore: switch CI to Github Actions

randi274:
### What does this PR do?
Removes CircleCI configuration, documentation, and tasks in favor of Github Actions.

**Update** - modified for the templates library specifically, now that this repository is limited to the singular package.

New Github Workflows:
- `test.yml` - Runs unit tests on every commit, except to main.
- `onPushToMain.yml` - On push to main, bumps the version and creates a tag and release in github.
- `onRelease.yml` - On successful creation of the release, publishes the library and the plugin to NPM.
- `failureNotification.yml` - if the previous two workflows fail, alert us in Slack using the IDEE webhook.
- `validatePR.yml` - ensure PRs have WIs associated.

@W-11596496@

### Notes
The first pass at this automation was created by the valet runner with #515. This was a good start, but it would have been all to easy for us to fall into the same pattern of building overly long singular jobs. Instead, fans of the Salesforce CLI may notice  that this change is shockingly similar to the new automation set forth in the plugin-analytics [repo](https://github.com/salesforcecli/plugin-analytics/pull/138). This is very much by design. Not only do we wish to keep our library in lockstep with the CLI, but the patterns established in that PR demonstrate a better understanding of the key attributes of github actions that we should also seek to leverage. The workflows are small, tight, and delegate a majority of the work to a shared space in the salesforcecli repository. If we find ourselves unable or unwilling to call those workflows directly, we would do well to create a shared forcedotcom actions library ourselves and continue to delegate the work separately.


test/service/templateService.test.ts
```diff
@@ -162,7 +162,7 @@ describe('TemplateService', () => {
         'LibraryCreateClass.cls'
       );
       const expectedApexClassContent =
-        'public with sharing class CustomLibraryCreateClass';
+        'public with sharing class LibraryCreateClass';
```
randi274, line 165: It was weird that it was failing, but I'm more surprised it worked before from reviewing the other tests here.

.github/workflows/onPushToMain.yml
```diff
@@ -0,0 +1,13 @@
...
+on:
+  push:
+    branches: [main]
+
+jobs:
+  release:
+    uses: salesforcecli/github-workflows/.github/workflows/githubRelease.yml@gbockus/updateReleaseToIncludeTokenInput
+    with:
+      ALT_GITHUB_TOKEN: ${{ secrets.IDEE_GH_TOKEN }}
+      ALT_GITHUB_USERNAME: svc-idee-bot
```
klewis-sfdc, line 12: Are username and email needed here?  I don't have them in this workflow in the apex repo.  Does this work using only the token?

.github/workflows/test.yml
```diff
@@ -0,0 +1,10 @@
+name: tests
+
+on:
+  push:
+    branches-ignore: [main]
+  workflow_dispatch:
+
+jobs:
+  unit-tests:
+    uses: salesforcecli/github-workflows/.github/workflows/unitTest.yml@randi/git-config-test-fix
```
klewis-sfdc, line 10: I see that your modification here does some git config for carriage returns.  Is this something that is only applicable for this repo?  Or is that something that should be configured for all of the repos for this workflow?

contributing/developing.md
```diff
@@ -47,8 +47,8 @@ Take a look at [CONTRIBUTING](../CONTRIBUTING.md) doc for making and merging pul
 
 Adding a new template:
 
-1. Define a new template type in `TemplateType`, and add available template options extending `TemplateOptions` in library [types](../packages/templates/src/utils/types.ts).
-2. Create a generator extending [`SfdxGenerator`](../packages/templates/src/generators/sfdxGenerator.ts) in [generators](../packages/templates/src/generators) folder. Take a look at [`ApexClassGenerator`](../packages/templates/src/generators/apexClassGenerator.ts) for an example.
+1. Define a new template type in `TemplateType`, and add available template options extending `TemplateOptions` in library [types](../src/utils/types.ts).
+2. Create a generator extending [`SfdxGenerator`](../src/generators/sfdxGenerator.ts) in [generators](../src/generators) folder. Take a look at [`ApexClassGenerator`](../src/generators/apexClassGenerator.ts) for an example.
```
klewis-sfdc, line 51: 👍🏽

klewis-sfdc (commented):
Great work, @randi274!  I left a few questions, but this is looking good to go!

.github/workflows/onPushToMain.yml
randi274: @klewis-sfdc good catch on this one! I had mentioned that we should add it [here](https://github.com/salesforcecli/github-workflows/pull/13#discussion_r1019255854), but that hasn't been done yet. I might see about updating Gordon's branch, since ideally we would be able to pass in all of this information.

.github/workflows/test.yml
randi274: This one was a bear to track down! That's a good idea though that this one would be worth putting up to a PR as well and getting merged in. Alternatively (or if that gets denied!) I could create a separate unit test yaml file for this repo, adding a note about why we needed a separate config for it.

src/templates/apexclass/DefaultApexClass.cls
```diff
@@ -2,4 +2,4 @@ public with sharing class <%= apiName %> {
     public <%= apiName %>() {
 
     }
-}
+}
```
randi274, line 5: I made a change here to fix an issue in vscode that was reported, since it seemed like it would fix some earlier test failures I was encountering. It turns out this was not necessary, so I'm going to pull this out into a subsequent PR for better tracking.

randi274: https://github.com/forcedotcom/salesforcedx-templates/commit/65fbada66e5ddda57541778b1210aef5b4595601

.github/workflows/onPushToMain.yml
randi274: Updating that PR also became a little more complex than I initially thought, so I reverted the extra fields for now!

.github/workflows/test.yml
klewis-sfdc: I like your idea of a separate unit test yaml file in this repo with the note for why it is needed.  I think that is the best option. And we should discuss as a team our longer term strategy for these scenarios (where one small difference in an otherwise shared workflow is needed)

randi274: Done! 🏁

.github/workflows/unitTestsWindows.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  windows-unit-tests:
+    strategy:
+      matrix:
+        node_version: [lts/-1, lts/*]
+      fail-fast: false
+    runs-on: windows-latest
+    steps:
+      - name: Git config check for carriage returns
+        run: git config --global core.autocrlf false
```
klewis-sfdc, line 13: 👍🏽

klewis-sfdc (approved):
Looks great @randi274!  Thanks for the changes, and a big thank you for getting these GitHub Actions in place!

PR #532, opened 2022-12-07
refactor: change the location of the default api version

randi274:
### What does this PR do?
Changes where we derive the default Salesforce API version. Previously we were conflating the version of the package with the Salesforce API version, so we could not benefit from semantic versioning. In order to keep our CI simple and benefit from semantic versioning, this will separate these two values. 

Going forward with the code as it is here, the only value that will need to change during major Salesforce version updates is the `salesforceApiVersion` in the `package.json`. The version should no longer be directly touched, as the `conventional-changelog` will update this as needed. 

A better longer term solution would be to dynamically generate this value, but this would require some understanding of what "default" should mean.

This PR also updates the `publishing.md` documentation to note this change and how publishing now works, and adds a manual release workflow to support the one-off publishing supported in other repos.

Follow-up to #527

### DIFF
--- a/.github/workflows/manualRelease.yml
+++ b/.github/workflows/manualRelease.yml
@@ -1,0 +1,36 @@
+name: manual release
+
+on:
+  workflow_dispatch:
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.IDEE_GH_TOKEN }}
+      - name: Conventional Changelog Action
+        id: changelog
+        uses: TriPSs/conventional-changelog-action@d360fad3a42feca6462f72c97c165d60a02d4bf2
+        # overriding some of the basic behaviors to just get the changelog
+        with:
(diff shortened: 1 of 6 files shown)


gbockus-sf (approved):
👍

test/generators/sfdxGenerator.test.ts
```diff
@@ -13,7 +13,7 @@ import { TemplateService } from '../../src/service/templateService';
 import * as YeomanEnvironment from 'yeoman-environment';
 
 describe('SfdxGenerator', () => {
-  const API_VERSION = '56.0';
+  const API_VERSION_STUB = '50';
```
randi274, line 16: I downgraded this to be super clear that this does not need to be changed with every release.

.github/workflows/manualRelease.yml
```diff
@@ -0,0 +1,36 @@
+name: manual release
```
gbockus-sf, line 1: total nit: could you capitalize both Manual and Release  so it's aligned with the other actions we have across repos?

PR #543, opened 2023-02-03
Sm/fixes

mshanemc:
### What does this PR do?
interface for TemplateOptions has apiVersion and outputDir as optional.  But SfdxGenerator is typed to make them required, so catch-22.

This moves those to instance props that are defaulted in the ctor.

unrelated:
`shelljs` is only used in some npm script, so it moved to devDeps
I bumped `tar` to help with an error related to node18 https://github.com/isaacs/node-tar/issues/321

needed for [@W-11976561@](https://gus.lightning.force.com/a07EE00001AE28uYAD) to remove a bunch of type-ignoring code

### DIFF
--- a/src/generators/projectGenerator.ts
+++ b/src/generators/projectGenerator.ts
@@ -55,15 +55,8 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
   }
 
   public writing(): void {
-    const {
-      projectname,
-      template,
-      defaultpackagedir,
-      manifest,
-      ns,
-      apiversion,
-      loginurl,
-    } = this.options;
+    const { projectname, template, defaultpackagedir, manifest, ns, loginurl } =
+      this.options;
     const folderlayout = [
       this.outputdir,
       projectname,
@@ -103,7 +96,7 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
         defaultpackagedir,
         namespace: ns,
         loginurl,
-        apiversion,
+        apiversion: this.apiversion,
         name: projectname,
       }
     );
@@ -114,7 +107,7 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
         this.destinationPath(
           path.join(this.outputdir, projectname, 'manifest', 'package.xml')
         ),
-        { apiversion }
+        { apiversion: this.apiversion }
       );
     }
 
--- a/src/generators/projectGenerator.ts
+++ b/src/generators/projectGenerator.ts
@@ -56,7 +56,6 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
 
   public writing(): void {
     const {
-      outputdir,
       projectname,
       template,
       defaultpackagedir,
@@ -66,7 +65,7 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
       loginurl,
     } = this.options;
     const folderlayout = [
-      outputdir,
+      this.outputdir,
       projectname,
       defaultpackagedir,
       'main',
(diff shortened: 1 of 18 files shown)


src/generators/visualforceComponentGenerator.ts
```diff
@@ -21,17 +21,18 @@ export default class VisualforceComponentGenerator extends SfdxGenerator<Visualf
   }
 
   public writing(): void {
-    const { template, outputdir, label, apiversion, componentname } =
-      this.options;
+    const { template, label, apiversion, componentname } = this.options;
```
randi274, line 24: Does apiversion here need to be removed?

src/generators/visualforceComponentGenerator.ts
```diff
@@ -21,17 +21,18 @@ export default class VisualforceComponentGenerator extends SfdxGenerator<Visualf
...
+      ),
       {}
     ),
       this.fs.copyTpl(
         this.templatePath('_component.component-meta.xml'),
         this.destinationPath(
-          path.join(outputdir, `${componentname}.component-meta.xml`)
+          path.join(this.outputdir, `${componentname}.component-meta.xml`)
         ),
         { vfLabel: label, apiVersion: apiversion }
```
randi274, line 37:
```suggestion
        { vfLabel: label, apiVersion: this.apiversion }
```

randi274: I wasn't seeing the VF component version set during testing, and think the `this` is all that's missing

src/generators/projectGenerator.ts
```diff
@@ -66,7 +65,7 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
       loginurl,
```
randi274, line 65: From the line above - does apiversion need to be cut here too?

src/generators/projectGenerator.ts
```diff
@@ -81,19 +80,24 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
...
       {}
     );
     this.fs.copyTpl(
       this.templatePath('sfdx-project.json'),
       this.destinationPath(
-        path.join(outputdir, projectname, 'sfdx-project.json')
+        path.join(this.outputdir, projectname, 'sfdx-project.json')
       ),
       {
         defaultpackagedir,
```
randi274, line 103: I can't comment on the apiversion below, but thinking that should also be `this.apiversion`. Is that what sets the sourceApiVersion? I'm not seeing that set when I create a new project with this template.

src/generators/projectGenerator.ts
```diff
@@ -108,7 +112,7 @@ export default class ProjectGenerator extends SfdxGenerator<ProjectOptions> {
       this.fs.copyTpl(
         this.templatePath(manifestFile),
         this.destinationPath(
-          path.join(outputdir, projectname, 'manifest', 'package.xml')
+          path.join(this.outputdir, projectname, 'manifest', 'package.xml')
         ),
         { apiversion }
```
randi274, line 117:
```suggestion
        { this.apiversion }
```

randi274: I'm also not seeing the <version> tag set in the manifest file, I think for ^^ reason

randi274 (changes requested):
QA Complete  ✔️  with VS Code - just a few changes!

With Custom template:
✅ Create apex class

With default template:
✅ Create apex class
✅ Create analytics template
✅ Create apex trigger
✅ Create lightning interface
✅ Create LWC
✅ Create LWC Test
✅ Create Aura cmp
✅ Create Aura app
✅ Create Aura event
✅ Create VF page

Failed:
❌ Create VF cmp -> no API version in the meta-xml
❌ Create Project - no sourceApiVersion in the sfdx-project.json
❌ Create Project w/ Manifest - no version set in the package.xml
^^ all should be solved with adding `this` to the apiversion.

src/generators/visualforceComponentGenerator.ts
mshanemc: yep!

randi274 (approved):
✅ Create VF cmp -> now setting API version in the meta-xml
✅ Create Project - now setting sourceApiVersion
✅ Create Project w/ Manifest - now setting version in the package.xml

PR #554, opened 2023-05-03
fix: eof newlines in templates

mshanemc:
### What does this PR do?
adds a newline for EOF for stuff under src/templates

normally prettier would do this, but this repo sets prettier to ignore src/templates.  :shrug:

https://github.com/forcedotcom/cli/issues/2107 @W-13147281@


mshanemc:
closed because old/ignored

PR #555, opened 2023-05-10
fix: update analytics template asset version

smithgp:
### What does this PR do?

Updates the `assetVersion` in the sample analytics app template template to 57.0, and updates the dashboard contained within that to the 57.0 format.

@W-13144541@


src/templates/analytics/waveTemplates/DefaultAnalyticsTemplate/dashboards/basicDashboard.json
```diff
@@ -69,15 +73,27 @@
...
-          "textColor": "rgb(9, 26, 62)"
+          "interactions": [],
+          "showActionMenu": true
         },
         "type": "text"
       }
     }
   }
-}
+}
```
selinaZhaoSF, line 99: nit: add an extra line, similar for the next file

selinaZhaoSF (approved):

src/templates/analytics/waveTemplates/DefaultAnalyticsTemplate/dashboards/basicDashboard.json
smithgp: Maybe? There's another PR #554 that's about the same thing. I think prettier might have (wrongly) rewritten these to remove the new line. I'll double check here.

smithgp: Yep, prettier was removing them, so I saved w/o formatting and the trailing newlines are back.

selinaZhaoSF:
I guess i did not have any write permission for this module, can u give me and Aaron (if needed) permission ? 
image

reidaelliott:
one other thing I was thinking in this repo is the scratch config.  I have been removing that mobile setting as it often fails to deploy.  Is that setting important to have for an analytics project?

reidaelliott (approved):

smithgp:
> one other thing I was thinking in this repo is the scratch config. I have been removing that mobile setting as it often fails to deploy. Is that setting important to have for an analytics project?

That's kind of unrelated to this change. But, I don't really care if we remove it, although it looks like the standard project template is still including that `enableS1EncryptedStoragePref2` false setting.

aarondailsf2023 (approved):

reidaelliott:
> > one other thing I was thinking in this repo is the scratch config. I have been removing that mobile setting as it often fails to deploy. Is that setting important to have for an analytics project?
> 
> That's kind of unrelated to this change. But, I don't really care if we remove it, although it looks like the standard project template is still including that `enableS1EncryptedStoragePref2` false setting.

yeah realize its unrelated, but in the same repo we don't touch often.  Your call maybe i'm the only one deleting that setting.

randi274 (approved):

PR #560, opened 2023-06-12
feat: upgrade node and salesforce/core to v4

randi274:
### What does this PR do?

Salesforce core has released a major version, so updated to keep in sync. Node also required updating, so I had it match what we're using in [apex-node](https://github.com/forcedotcom/salesforcedx-apex/blob/main/package.json). We also need to bump a major version anyway since Salesforce is at v58 now.

BREAKING CHANGE: Node v16.13 is now required.

@W-13577556@

### Testing Notes:
- To set the github repo location: https://developer.salesforce.com/tools/vscode/en/user-guide/byotemplate#set-default-template-location. For example, I ran this command (which may work for QA too). `sfdx config:set customOrgMetadataTemplates=https://github.com/randi274/salesforcedx-templates/tree/main/src/templates`


CristiCanizales (approved):
LGTM :) I verified it compiles successfully and the commands still work as expected :)

:shipit:

peternhale (approved):

PR #569, opened 2023-09-20
Update slack notification workflow name

CristiCanizales:
Updates title and file name for the slack notification workflow

[skip-validate-pr]


daphne-sfdc (approved):
Name change approved because the workflow is also called on success use cases. 🙂

PR #570, opened 2023-11-10
fix: upgrade min node engine to 18

peternhale:
### What does this PR do?
Upgrades minimum node engine to 18
Bumps typescript from 4 to 5
Bumps a number of other dependencies w/o introducing breaking changes

@W-14196509@


package.json
```diff
@@ -13,58 +13,59 @@
...
-    "ts-node": "^10.0.0",
-    "typescript": "^4.4.4",
+    "sinon": "^17.0.1",
+    "ts-node": "^10.4.0",
+    "typescript": "^5.2.2",
     "yeoman-assert": "^3.1.1"
   },
   "engines": {
-    "node": ">=16.13.0"
+    "node": ">=18.18.2"
```
smithgp, line 68: Could you bump the `.nvmrc` to match as well?

smithgp: Also, looks like @mshanemc is doing the node bump in #572 as well, probably want to coordinate.

daphne-sfdc (approved):
The CLI Integration and Apex extensions are both working as expected after this Node 18 upgrade.  PR approved! ✅

QE doc here: https://salesforce.quip.com/HIO6AnnVbN1l

PR #572, opened 2023-11-14
feat!: bump sfdx-core to 6, engines to node 18+

mshanemc:
### What does this PR do?
use core v6
https://github.com/forcedotcom/sfdx-core/blob/main/MIGRATING_V5-V6.md
require node 18+(via pjson engines)

@W-14085765@


package.json
```diff
@@ -64,7 +64,7 @@
     "yeoman-assert": "^3.1.1"
   },
   "engines": {
-    "node": ">=16.13.0"
+    "node": ">=18.0.0"
```
smithgp, line 67: Can you bump the `.nvmrc` file, too?

mshanemc: done.  never used that one before

peternhale (approved):

PR #574, opened 2023-12-07
fix: bump core to 5.3.20

peternhale:
### What does this PR do?
bumps salesforce core to 5.3.20
@W-11622957@


mingxuanzhangsfdx (approved):
I pulled the change and built locally. Looks good.

PR #575, opened 2024-01-05
feat: add new empty apex unit test template

WillieRuemmele:
### What does this PR do?

adds an empty apex unit test file template to be used for E4D

@W-14709113@


gbockus-sf (approved):
🚢

PR #581, opened 2024-04-10
fix: update salesforce/core

mshanemc:
### What does this PR do?
sfdx-core is only used to import `Global` so this shouldn't be a breaking change (no Connection, testSetup, SfError, etc).  See https://github.com/forcedotcom/sfdx-core/blob/main/MIGRATING_V6-V7.md for change info

@W-15387805@

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@salesforce/templates",
-  "version": "60.1.1",
+  "version": "61.0.0-beta.0",
   "salesforceApiVersion": "59",
   "description": "Salesforce JS library for templates",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@salesforce/templates",
-  "version": "60.1.1",
+  "version": "61.0.0-beta.0",
```
peternhale, line 3: @mshanemc this follows core API version numbers, so best to use 60.2.0-beta.0.

peternhale (changes requested):

peternhale (approved):

PR #587, opened 2024-06-19
feat: remove dependency on yeoman

mdonnalley:
### What does this PR do?

Removes dependency on yeoman-generator and yeoman-environment

@W-16025380@


src/service/templateService.ts
```diff
@@ -5,30 +5,29 @@
...
-
-interface FsError extends Error {
-  code: string;
+import {
+  type CreateOutput,
+  type TemplateOptions,
+  TemplateType,
+} from '../utils/types';
+
+async function importGenerator(templateType: TemplateType) {
```
mshanemc, line 14: thank you for not making a Factory !

src/service/templateService.ts
```diff
@@ -5,30 +5,29 @@
...
+  const generatorClass =
```
mshanemc, line 15: given this pattern, there should probably be something (could be a UT) that enforces the naming convention and location for the generator classes?

mshanemc: ex: you could iterate TemplateType and check that the corresponding generator exists.

an alternative design could be changing TemplateType to be a Map that connected each TemplateType to its corresponding generator to prevent error.

[making these dx comments because the generator stuff is kinda meant to be contributed by lots of different owners]

test/service/templateService.test.ts
```diff
@@ -330,17 +354,14 @@ describe('TemplateService', () => {
...
       }
       const customTemplates = TEST_CUSTOM_TEMPLATES_REPO;
 
-      await templateService.setCustomTemplatesRootPathOrGitRepo(
-        customTemplates
-      );
+      await setCustomTemplatesRootPathOrGitRepo(customTemplates);
 
-      sinonAssert.calledOnce(streamStub);
+      assert.calledOnce(streamStub);
```
mshanemc, line 364: if it fails, does that mean it was called 0 or more than once?
I prefer `callCount` over these.  If I expect `1` but it fails, it tells me how many times it was called.

src/utils/logger.ts
mshanemc, line 1: this is awesome!

src/generators/sfGenerator.ts
mshanemc, line 1: I understand sfdx=>sf for the CLI's stuff.  This lib is used by CLI, but shared by more of the entire salesforce dx.

this class isn't exported, so it could be less "branded" and more about what it's to be used for in this library.  (ex: baseGenerator because every real generator should extend it)

mshanemc (changes requested):
our plugin that uses this has pretty great NUT coverage (real fs across OSes, etc).  I'd love to see those tests run for library changes to enable ongoing refactoring and dep maintenance for this lib.

If the lib owners aren't up for that, I'd like to see a run of them (maybe branch of a branch) to verify these changes don't affect output.

src/generators/sfGenerator.ts
```diff
@@ -0,0 +1,278 @@
...
+
+import * as fs from 'fs';
+import { access, mkdir, writeFile, readFile } from 'node:fs/promises';
+import * as path from 'path';
+import { CreateOutput, TemplateOptions } from '../utils/types';
+import { renderFile } from 'ejs';
+import { nls } from '../i18n';
+import { loadCustomTemplatesGitRepo } from '../service/gitRepoUtils';
+
+async function pathExists(path: string): Promise<boolean> {
```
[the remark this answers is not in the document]
mdonnalley: Just blindly copying fs-extra since that's what yeoman was using. I don't think it's actually necessary so I'll remove it

mshanemc:
I still see one failing NUT on the plugin when I use this branch and your branch on the plugin (running locally)

1) TemplateCommand
       output
         should log json output when flag is specified:
     AssertionError: expected [ 'foo.cls', 'foo.cls-meta.xml' ] to be empty
      at Context.<anonymous> (file:///Users/shane.mclaughlin/eng/salesforcecli/plugin-templates/test/utils/templateCommand.nut.ts:41:61)
      at processImmediate (node:internal/timers:478:21)

test/service/templateService.test.ts
```diff
@@ -481,4 +477,436 @@ describe('TemplateService', () => {
       }
     });
   });
+
+  describe('Generators', () => {
+    it('should have a generator for each TemplateType', async () => {
```
mshanemc, line 482: cool!

package.json
```diff
@@ -52,17 +49,17 @@
     "eslint-plugin-import": "^2.29.0",
     "eslint-plugin-jsdoc": "^46.8.2",
     "eslint-plugin-prettier": "^3.1.3",
+    "fs-extra": "^11.2.0",
```
mshanemc, line 52: I don't see why fs-extra is needed.

src/generators/sfGenerator.ts
```diff
@@ -0,0 +1,278 @@
...
+      );
+    }
+  }
+
+  public async run(opts?: {
+    cwd?: string;
+    customTemplatesRootPathOrGitRepo?: string;
+    sourceRootPartial?: string;
+  }): Promise<CreateOutput> {
+    const cwd = opts?.cwd ?? process.cwd();
```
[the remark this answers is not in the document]
peternhale: Should be adequate.

peternhale (approved):

mshanemc (approved):

PR #590, opened 2024-07-09
feat: add typescript template for lwc

jmsjtu:
### What does this PR do?

This PR adds a new TypeScript template for LWCs.

@W-15696151@


src/generators/lightningComponentGenerator.ts
```diff
@@ -158,7 +159,7 @@ export default class LightningComponentGenerator extends BaseGenerator<Lightning
           path.join(
             this.outputdir,
             camelCaseComponentName,
-            `${camelCaseComponentName}.js`
+            `${camelCaseComponentName}.${ext}`
```
jmsjtu, line 162: I ran into some issues with the repo containing `.ts` files in the template, I opted to use a `.js` extension for in the template since there's nothing specific to TypeScript in these files.

test/service/templateService.test.ts
```diff
@@ -638,6 +638,34 @@ describe('TemplateService', () => {
...
+      );
+
+      chai
+        .expect(result.created.sort())
+        .to.deep.equal(
+          [
+            'testsoutput/libraryCreate/lwc/libraryCreateComponent/libraryCreateComponent.ts',
+            'testsoutput/libraryCreate/lwc/libraryCreateComponent/libraryCreateComponent.html',
+            'testsoutput/libraryCreate/lwc/libraryCreateComponent/__tests__/libraryCreateComponent.test.ts',
+            'testsoutput/libraryCreate/lwc/libraryCreateComponent/libraryCreateComponent.js-meta.xml',
```
mingxuanzhangsfdx, line 661: I noticed when the typescript LWC is created, there will always be an error shown up at the start of the file in VS Code - `No base file for markup://c:asdLightningComponentBundle` since the js file is missing and it makes the `js-meta.xml` file red until the typescript component is compiled. Is that what we want?

src/generators/lightningComponentGenerator.ts
mingxuanzhangsfdx: Wondering what kind of issues you ran into

src/templates/lightningcomponent/lwc/typeScript/__tests__/typeScript.test.js
```diff
@@ -0,0 +1,25 @@
+import { createElement } from 'lwc';
```
mingxuanzhangsfdx, line 1: Correct me if I am wrong - It looks to me that all the newly created templates are the same as them in the default template. Then why not reuse them?

src/generators/lightningComponentGenerator.ts
jmsjtu: It looks like the repo is configured to search for all `.ts` file extensions and do the following:

1. Lint the components
2. Run type validation

Both of these will fail because the placeholders in the source templates. 

I wasn't sure if ignore the file paths would have any unintended side effects in the repo so I changed these to `.js` files.

jmsjtu: @mingxuanzhangsfdx on second thought, I made changes to the repo to ignore the specific TS files and changed them back to `.ts` file extensions.

see - 5c2f0ce29be89e05ffcb632d52e37a46f35842fa

src/templates/lightningcomponent/lwc/typeScript/__tests__/typeScript.test.js
jmsjtu: @mingxuanzhangsfdx I updated the file extensions to be `.ts` files, see my comment [here](https://github.com/forcedotcom/salesforcedx-templates/pull/590#discussion_r1681793131).

I wanted to start off with the templates being identical and over time we could add some more specific TypeScript related syntax to it.

We can reuse the existing `.js` template as well, but I figured might as well create the `.ts` template while we're at it.

Let me know what you think I can do it either way.

mingxuanzhangsfdx (approved):
Change looks good to me! I am able to generate the typescript lwc with the new template and javascript template is created properly as well.

PR #592, opened 2024-07-30
feat: remove core from dependencies

mingxuanzhangsfdx:
### What does this PR do?
This PR removes `@salesforce/core` from dependencies, reduces the total size of the library.

@W-16301500@

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,10 +6,7 @@
     "sourceType": "module"
   },
   "plugins": ["@typescript-eslint", "jsdoc", "eslint-plugin-header"],
-  "extends": [
-    "plugin:@typescript-eslint/recommended",
-    "plugin:prettier/recommended"
-  ],
+  "extends": ["plugin:@typescript-eslint/recommended"],
   "rules": {
     "ban-ts-ignore": "off",
     "@typescript-eslint/ban-ts-ignore": "off",
--- a/src/utils/createUtil.ts
+++ b/src/utils/createUtil.ts
@@ -6,14 +6,11 @@
  */
 import * as fs from 'fs';
 import * as path from 'path';
-import * as os from 'node:os';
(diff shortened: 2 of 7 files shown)


.eslintrc.json
```diff
@@ -6,10 +6,7 @@
     "sourceType": "module"
   },
   "plugins": ["@typescript-eslint", "jsdoc", "eslint-plugin-header"],
-  "extends": [
-    "plugin:@typescript-eslint/recommended",
-    "plugin:prettier/recommended"
```
mingxuanzhangsfdx, line 11: I found out that with the "plugin:prettier/recommended" there will be a conflict of linting with auto-save and prettier. Given that it will lint with prettier when push, there is no need to add this check with conflict.

src/utils/createUtil.ts
```diff
@@ -6,11 +6,14 @@
  */
 import * as fs from 'fs';
 import * as path from 'path';
+import * as os from 'node:os';
 import { nls } from '../i18n';
 
 /* tslint:disable:no-unused-expression */
 
 export class CreateUtil {
+  private static SFDX_STATE_FOLDER = '.sfdx';
```
peternhale, line 15: I would prefer the logic for DIR be moved to `src/service/gitRepoUtils.ts`. It is only used in one place.

src/utils/createUtil.ts
```diff
@@ -75,4 +78,11 @@ export class CreateUtil {
     }
     return subdirs;
   }
+
+  /**
+   * The full system path to the preferred global state folder
+   */
+  public static get DIR(): string {
```
peternhale, line 85: move to src/service/gitRepoUtils.ts

peternhale (changes requested):

peternhale (approved):

PR #593, opened 2024-07-31
chore: make templates bundle-able

mingxuanzhangsfdx:
### What does this PR do?
This PR removes the friction that blocks templates from bundling.
1. transition the generator imports to static imports.
2. remove the templates-bundle publishing workflow since templates will be bundle-able.

@W-15736930@


scripts/esbuild.config.js
```diff
@@ -0,0 +1,41 @@
+const { build } = require('esbuild');
+const { exec } = require('child_process');
+const { Generator } = require('npm-dts');
+new Generator({
+  output: `dist/src/index.d.ts`,
+}).generate();
+
+const sharedConfig = {
+  bundle: true,
+  format: 'cjs',
```
gbockus-sf, line 10: can the cli team consume a cjs module now that they are an esm shop?  Might influence what we publish.

src/service/templateService.ts
```diff
@@ -64,9 +101,9 @@ export class TemplateService {
    * @param templateOptions template options
    * @param customTemplatesRootPathOrGitRepo custom templates root path or git repo. If not specified, use built-in templates
    */
-  public async create<TOptions extends TemplateOptions>(
+  public async create(
     templateType: TemplateType,
-    templateOptions: TOptions,
+    templateOptions: any,
```
gbockus-sf, line 106: nice

src/service/templateService.ts
```diff
@@ -5,18 +5,55 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import {
-  type CreateOutput,
-  type TemplateOptions,
-  TemplateType,
-} from '../utils/types';
+import { type CreateOutput, TemplateType } from '../utils/types';
```
peternhale, line 8: Please move the imports/Generators/map to the generators folder.

src/service/templateService.ts
```diff
@@ -5,18 +5,55 @@
...
+  ['lightningEventGenerator', lightningEventGenerator],
+  ['lightningInterfaceGenerator', lightningInterfaceGenerator],
+  ['lightningTestGenerator', lightningTestGenerator],
+  ['projectGenerator', projectGenerator],
+  ['staticResourceGenerator', staticResourceGenerator],
+  ['visualforceComponentGenerator', visualforceComponentGenerator],
+  ['visualforcePageGenerator', visualforcePageGenerator],
+]);
 
 export async function importGenerator(templateType: TemplateType) {
```
peternhale, line 51: The contents of this function no longer returns a promise.

peternhale (changes requested):

scripts/esbuild.config.js
mingxuanzhangsfdx: That's a good question. I will have a try.

package.json
```diff
@@ -4,12 +4,12 @@
...
   "description": "Salesforce JS library for templates",
   "bugs": "https://github.com/forcedotcom/salesforcedx-templates/issues",
-  "main": "lib/index.js",
+  "main": "dist/src/index.js",
   "author": "Salesforce",
   "homepage": "https://github.com/forcedotcom/salesforcedx-templates",
   "license": "BSD-3-Clause",
   "files": [
-    "/lib",
+    "/dist",
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: If we are going to use bundle to publish the project, I guess we have to change it because bundled artifacts exist in dist folder.

scripts/esbuild.config.js
mingxuanzhangsfdx: Since we do not need to bundle the project here. I think we do not need to take care of the issue.

peternhale (approved):

peternhale:
QA Notes:

Linked this module to plugin-templates and smoke tested the template commands.

- Commands that only have one template (project)
- Commands that have multiple types (Apex class)
-   Used Default
-   Used named

Smoke tests worked fine.

PR #598, opened 2024-08-26
fix: move default api version to a constant

peternhale:
@W-16407766@
remove the constants.json file in favor of exported const for api version

### DIFF
--- a/test/generators/baseGenerator.test.ts
+++ b/test/generators/baseGenerator.test.ts
@@ -60,12 +60,3 @@ describe('BaseGenerator', () => {
     expect(validateOptionsStub.calledOnce).to.be.true;
   });
 });
-
-describe('getDefaultApiVersion', () => {
-  it('should return the default api version', async () => {
-    const constants = await import('../../src/utils/constants.json');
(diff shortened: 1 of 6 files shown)


test/generators/baseGenerator.test.ts
```diff
@@ -60,12 +60,3 @@ describe('BaseGenerator', () => {
     expect(validateOptionsStub.calledOnce).to.be.true;
   });
 });
-
-describe('getDefaultApiVersion', () => {
```
CristinaCanizales, line 64: Why delete this test?

yarn.lock
CristiCanizales, line 1: was this file supposed to be included? I don't see any changes in package.json

test/generators/baseGenerator.test.ts
peternhale: It is just a constant now, no magic to get it loaded from json.

yarn.lock
peternhale: The change is due to a sinon module change, indirect dependency for tests

CristiCanizales (approved):

PR #660, opened 2025-04-14
Revert 657 dependabot/npm and yarn/src/templates/project/lwc/eslint plugin lwc 3.0.0

daphne-sfdc:
Reverts https://github.com/forcedotcom/salesforcedx-templates/pull/657

### DIFF
--- a/src/templates/project/package.json
+++ b/src/templates/project/package.json
@@ -16,7 +16,7 @@
     "precommit": "lint-staged"
   },
   "devDependencies": {
-    "@lwc/eslint-plugin-lwc": "^3.0.0",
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -9,6 +9,8 @@ updates:
     ignore:
       - dependency-name: 'eslint'
```
mshanemc, line 10: why not remove the redundant eslint item?

mshanemc (approved):
approved, 1 non-blocking comment

.github/dependabot.yml
daphne-sfdc: Good idea; it's now removed in both places.

PR #669, opened 2025-06-16
feat: add ESLint 9 support to template (#619)

pozil:
### What does this PR do?
This PR adds support for ESLint 9 in project template as requested in #619.

Changes:
- updated project template dependencies to the latest versions
- got rid of unneeded dependencies: `@lwc/engine-dom`, `@lwc/jest-preset` and `jest` as these are already provided by `@salesforce/sfdx-lwc-jest`. See sample apps best practices.
- migrated and regrouped aura/lwc config into a single ESLint 9 config file at project root level
- fixed a Husky warning related to the `.husky/pre-commit` extra lines

### What issues does this PR fix or reference?
Issue: #619

PR #672, opened 2025-06-22
fix: run husky setup after npm install

nrakuyama:
### What does this PR do?

This PR ensures that Husky is set up automatically after running npm install.

Before the Husky version in the template was upgraded to v9, Git hooks were set up via `husky install` in the postinstall script. However, that script was changed in #662 and removed in #669. This change restores the behavior by reintroducing the setup step.
Note that `husky init` has been avoided because it overwrites the contents of existing `.husky/pre-commit`. Instead, this PR runs `husky` to setup only Git hooks.

Additionally, it changes that Husky fails clearly if it encounters an error.

#662 #669


nrakuyama:
Hi, @pozil @mshanemc.
I also encountered the Husky setup issue that addressed in #669.
This PR restores the behavior from before #662, where Git hooks were set up after running `npm install`.
Could you check this PR?

pozil:
@nrakuyama I'm not a maintainer on this project but I made some contributions based on our experience with sample apps.
We switched to `prepare` instead of `precommit` based on the husky [official recommendations](https://typicode.github.io/husky/get-started.html#husky-init-recommended). Here's a [detailed explanation](https://github.com/typicode/husky/issues/884) about why they made the change.

In any case, you should not manually run `npm run prepare` as it's automatically run as part of [npm lifecycle scripts](https://docs.npmjs.com/cli/v11/using-npm/scripts#life-cycle-scripts).

nrakuyama:
@pozil 
Thank you for checking and correcting my comment.
I understood that the `prepare` script runs in the development mode of `npm install`. I apologize for the misunderstanding.
I will close this PR.

PR #679, opened 2025-08-05
fix(deps): dependabot updates that hadn't automerged

mshanemc:
### What does this PR do?

consolidate and test dependabot PRs
- https://github.com/forcedotcom/salesforcedx-templates/pull/678
- https://github.com/forcedotcom/salesforcedx-templates/pull/524
- https://github.com/forcedotcom/salesforcedx-templates/pull/544

moves to current version of `tar` library, which is a bit different API-wise.  It's not very compatible with streams, so there's now a temp file involved

[skip-validate-pr]


test/service/templateService.test.ts
```diff
@@ -343,52 +343,44 @@ describe('TemplateService', () => {
...
-      await setCustomTemplatesRootPathOrGitRepo(
+      const result = await setCustomTemplatesRootPathOrGitRepo(
         customTemplates,
         forceLoadingRemoteRepo
       );
 
-      sinon.assert.calledOnce(streamStub);
+      // Verify that the function returns a path and the folder exists
+      assert(typeof result === 'string');
+      assert(fs.existsSync(result));
```
daphne-sfdc, line 383: To be consistent with Line 353.

```suggestion
      chai.expect(fs.existsSync(result)).to.be.true;
```

daphne-sfdc (approved):
Just a nit in the unit test 🙂

PR #821, opened 2026-05-22
feat: add lightning embedding template generator

aman06it:
## Summary
- Adds `LightningEmbeddingGenerator` with default LWC bundle templates (html/js/css/meta), i18n strings, and `LightningEmbeddingOptions`.
- Adds `TemplateType.LightningEmbedding` enum value and registers the generator in the type-to-generator map.
- Template emits `<lightning-embedding>` (the base component the wrapper consumes).
- Validates `--src` URL (https only; `http://localhost` and `http://127.0.0.1` permitted for local development), `--shell-title` (non-empty), and `--sandbox` tokens against the W3C-defined set.
- Per team alignment in #plugin-design-templates: name aligns with the `<lightning-embedding>` base component (avoids both `mfe` and `microfrontend`).
- Consumed by the new `sf template generate lightning embedding` command in salesforcecli/plugin-templates#943.


peternhale (commented):
## Review

Good addition — follows existing generator patterns well, solid test coverage, and clean export surface for the plugin consumer.

Two actionable items and one nit:

### 1. Validate `src` and `shellTitle` for string-breaking characters

A valid URL like `https://example.com/path?q='foo` or a title like `My "App"` would produce a broken `.js` or `.html` file because the EJS templates use raw `<%= %>` interpolation into a JS string literal (single-quoted) and an HTML attribute (double-quoted) respectively.

Simplest fix: reject `'` in `src` and `"` in `shellTitle` during `validateOptions()`. These characters are technically legal but practically never needed, and it keeps the template rendering safe without adding an escaping layer.

### 2. Make `internal` required in `LightningEmbeddingOptions`

Every other `*Options` interface with `internal` declares it as a required `boolean`. The optional `?` is inconsistent and the generator already uses it in a truthy check without guarding against `undefined`. Low effort, removes ambiguity for callers.

### 3. (Nit) Update the PR title

Still says "microfrontend" — should say "lightning embedding" to match the code after the rename in `99828dc`.

peternhale (changes requested):
Requesting changes for the two items noted in the review comment: (1) validate src/shellTitle for string-breaking characters to prevent malformed generated files, and (2) make internal required in LightningEmbeddingOptions for consistency.

aman06it:
> Requesting changes for the two items noted in the review comment: (1) validate src/shellTitle for string-breaking characters to prevent malformed generated files, and (2) make internal required in LightningEmbeddingOptions for consistency.

Your suggested changes have been incorporated.

peternhale (approved):

Issue #848, opened 2026-06-18
Add templates for commonly used Apex class types


mitchspano:
**Is your feature request related to a problem? Please describe.**

When creating an Apex class from the VS Code command palette, the user is prompted to pick a class type. The current options (`DefaultApexClass`, `ApexException`, `InboundEmailService`) don't cover the patterns developers reach for most often, so the template rarely saves the developer any effort.

**Describe the solution you'd like**
Add templates for common asynchronous Apex patterns, starting with:
- **Queueable** : implements `Queueable` (+ `Finalizer` with a job-status failure check).
- **Batchable** : implements `Database.Batchable`, with a parameterized SObject type (defaults to `SObject`).

**Describe alternatives you've considered**
Leaving it as-is: developers copy/paste boilerplate from docs each time.

**Additional context**
Initial Queueable and Batchable templates plus tests are already drafted. Unsure if we want to tackle any other additional templates right away.

PR #853, opened 2026-06-21
feat: add apex templates for common apex class types

mitchspano:
### What does this PR do?

Adds two Apex class templates: **Batchable** and **Queueable**.  The goal is to scaffold these classes *the right way* so the user (or coding agent) has a starting point already follows current best practices.

#### What's included
- **`Batchable.cls`**: strongly typed `Database.Batchable<SObject>` with a typed `List<SObject>` scope, inline (compile-checked) SOQL instead of a dynamic string query, and `Database.RaisesPlatformEvents` so batch iteration failures surface as platform events.
- **`Queueable.cls`**: implements `Finalizer` and attaches it, so post-job logic runs reliably even after an unhandled exception; `UNHANDLED_EXCEPTION` is handled explicitly.
- New optional `sobjecttype` option on `ApexClassOptions` (defaults to `SObject`), the two new `template` values, and tests.

#### Rollout : This is the first of two PRs
1. **This PR (`salesforcedx-templates`):** the raw templates. No user-facing surface yet.
2. **Follow-up (`salesforcedx-vscode`):** enable them in the Command Palette (`SFDX: Create Apex Class`) entries, localized descriptions, and an SObject prompt for Batchable.

#### Open question: Do we want to make the library hermetic?
A template today is just a `.cls` with no description, labels, i18n, or declared inputs.  Every consumer hardcodes that info, which is why this takes two PRs and the repos can drift.

**Is there any appetite to co-locate each template's content + metadata + i18n** as the single source of truth in the templates repository? Then one PR here would propagate everywhere: label, description, translations, and required inputs included. I know there have been some challenges choreographing sequences of changes which rely on templates lib; perhaps this could alleviate some future headaches.

#848


mshanemc:
Thanks for the contribution! Adopting as #860 so our internal CI + AI review pipeline can run (forks have limited access).

Your commits preserved with you as Author. Follow #860 for status — we'll merge from there.

GUS WI: W-23219494

PR #860, opened 2026-06-28
feat: add apex templates for common apex class types - W-23219494

mshanemc:
### What does this PR do?

Adds two Apex class templates: **Batchable** and **Queueable**.  The goal is to scaffold these classes *the right way* so the user (or coding agent) has a starting point already follows current best practices.

#### What's included
- **`Batchable.cls`**: strongly typed `Database.Batchable<SObject>` with a typed `List<SObject>` scope, inline (compile-checked) SOQL instead of a dynamic string query, and `Database.RaisesPlatformEvents` so batch iteration failures surface as platform events.
- **`Queueable.cls`**: implements `Finalizer` and attaches it, so post-job logic runs reliably even after an unhandled exception; `UNHANDLED_EXCEPTION` is handled explicitly.
- New optional `sobjecttype` option on `ApexClassOptions` (defaults to `SObject`), the two new `template` values, and tests.

#### Rollout : This is the first of two PRs
1. **This PR (`salesforcedx-templates`):** the raw templates. No user-facing surface yet.
2. **Follow-up (`salesforcedx-vscode`):** enable them in the Command Palette (`SFDX: Create Apex Class`) entries, localized descriptions, and an SObject prompt for Batchable.

#### Open question: Do we want to make the library hermetic?
A template today is just a `.cls` with no description, labels, i18n, or declared inputs.  Every consumer hardcodes that info, which is why this takes two PRs and the repos can drift.

**Is there any appetite to co-locate each template's content + metadata + i18n** as the single source of truth in the templates repository? Then one PR here would propagate everywhere: label, description, translations, and required inputs included. I know there have been some challenges choreographing sequences of changes which rely on templates lib; perhaps this could alleviate some future headaches.

#848

---

**Adopted from #853** by @mitchspano — commits preserved with author credit. Replayed onto main for internal CI / AI review.

### Reviewer notes

**medium** `Batchable.cls:4` — `SELECT Id FROM SObject` has no WHERE clause; generates a query over all records. Added TODO comment to prompt developer to add a filter.

**medium** `Queueable.cls:1` — implementing `Finalizer` on same class loses instance state between transactions (finalizer runs in a separate transaction). Consider extracting to a separate class for stateful patterns.

**medium** `Batchable.cls:1` — `with sharing` on async classes applies the enqueueing user's sharing rules, which can silently restrict batch data access in scheduled/system contexts. Worth documenting in generated comments for awareness.

**medium** `src/utils/types.ts` — `sobjecttype?: string` is valid for all templates but only consumed by `Batchable`. No runtime error for other templates, but a discriminated union would make intent clearer. **Resolved:** `ApexClassOptions` is now a discriminated union — `sobjecttype?: never` on non-Batchable, `?: string` only on `Batchable`.

**medium** `apexClassGenerator.ts` — empty-string `sobjecttype` bypassed `??` and rendered invalid Apex (`Database.Batchable<>`). **Resolved:** generator uses `sobjecttype?.trim() || 'SObject'`, defaulting empty/whitespace to `SObject`.

**medium** `apexClassGenerator.ts` — `sobjecttype` was unvalidated, allowing invalid Apex identifiers (`'My Object'`, `'123Invalid'`). **Resolved:** added `CreateUtil.checkSObjectType`, called from `validateOptions` for `Batchable`. Did not reuse `checkInputs` (rejects `__`, would break valid custom objects like `My_Object__c`); new validator allows custom/namespaced objects while rejecting bad identifiers.

@W-23219494@

Co-authored-by: Mitch Spano <mitch@sre.ai>

### DIFF
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -123,19 +123,27 @@ export interface AnalyticsTemplateOptions extends TemplateOptions {
   templatename: string;
 }
 
-export interface ApexClassOptions extends TemplateOptions {
+type ApexClassBaseOptions = TemplateOptions & { classname: string };
+
+type NonBatchableApexClassOptions = ApexClassBaseOptions & {
   template:
     | 'DefaultApexClass'
     | 'BasicUnitTest'
     | 'ApexUnitTest'
     | 'ApexException'
     | 'InboundEmailService'
-    | 'Queueable'
-    | 'Batchable';
-  classname: string;
-  /** SObject type for the Batchable template. Defaults to 'SObject'. */
+    | 'Queueable';
+  sobjecttype?: never;
+};
+
+type BatchableApexClassOptions = ApexClassBaseOptions & {
+  template: 'Batchable';
   sobjecttype?: string;
-}
+};
+
+export type ApexClassOptions =
+  | NonBatchableApexClassOptions
+  | BatchableApexClassOptions;
 
 type ApexTriggerEvent =
   | 'before insert'
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -129,8 +129,12 @@ export interface ApexClassOptions extends TemplateOptions {
     | 'BasicUnitTest'
     | 'ApexUnitTest'
     | 'ApexException'
-    | 'InboundEmailService';
+    | 'InboundEmailService'
+    | 'Queueable'
+    | 'Batchable';
   classname: string;
+  /** SObject type for the Batchable template. Defaults to 'SObject'. */
+  sobjecttype?: string;
 }
 
 type ApexTriggerEvent =
--- a/src/generators/apexClassGenerator.ts
+++ b/src/generators/apexClassGenerator.ts
@@ -13,6 +13,12 @@ export default class ApexClassGenerator extends BaseGenerator<ApexClassOptions> 
   public validateOptions(): void {
     CreateUtil.checkInputs(this.options.template);
     CreateUtil.checkInputs(this.options.classname);
+    if (
+      this.options.template === 'Batchable' &&
+      this.options.sobjecttype?.trim()
+    ) {
+      CreateUtil.checkSObjectType(this.options.sobjecttype.trim());
+    }
   }
 
   public async generate(): Promise<void> {
@@ -25,7 +31,7 @@ export default class ApexClassGenerator extends BaseGenerator<ApexClassOptions> 
       {
         apiName: classname,
         ...(template === 'Batchable' && {
-          sobjectType: sobjecttype ?? 'SObject',
+          sobjectType: sobjecttype?.trim() || 'SObject',
         }),
       }
     );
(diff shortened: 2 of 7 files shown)


src/utils/types.ts
```diff
@@ -129,8 +129,12 @@ export interface ApexClassOptions extends TemplateOptions {
     | 'BasicUnitTest'
     | 'ApexUnitTest'
     | 'ApexException'
-    | 'InboundEmailService';
+    | 'InboundEmailService'
+    | 'Queueable'
+    | 'Batchable';
   classname: string;
+  /** SObject type for the Batchable template. Defaults to 'SObject'. */
+  sobjecttype?: string;
```
mshanemc, line 141: this should only be available with Batchable, it should be never on the other cases.

peternhale (changes requested):
## Review Findings (Medium Effort)

Found **2 validation bugs** that could generate invalid Apex code:

### 1. Empty string `sobjecttype` bypasses nullish coalescing
**Location:** `src/generators/apexClassGenerator.ts:28`

The `??` operator only catches `null`/`undefined`, not empty string. When `sobjecttype: ''` is passed:
```typescript
sobjectType: sobjecttype ?? 'SObject'  // evaluates to ''
```

This causes the Batchable template to render invalid Apex:
- `Database.Batchable<>` (missing type)
- `SELECT Id FROM` (missing object)  
- `List<>` (missing type)

**Fix:** Use `sobjecttype?.trim() || 'SObject'` or validate/reject empty strings in `validateOptions()`.

### 2. Missing validation for `sobjecttype` parameter
**Location:** `src/generators/apexClassGenerator.ts:13-16`

The `validateOptions()` method validates `template` and `classname` using `CreateUtil.checkInputs()`, but `sobjecttype` is never validated before being used in templates.

This allows invalid Apex identifiers:
- `sobjecttype: 'My Object'` → `Database.Batchable<My Object>` (spaces invalid)
- `sobjecttype: '123Invalid'` → `Database.Batchable<123Invalid>` (must start with letter)

Both fail Apex compilation.

**Fix:** Add validation in `validateOptions()`:
```typescript
if (this.options.template === 'Batchable' && this.options.sobjecttype) {
  CreateUtil.checkInputs(this.options.sobjecttype);
}
```

---

**Recommendation:** Add proper validation for the `sobjecttype` parameter to prevent invalid Apex code generation.

src/utils/types.ts
mshanemc: Addressed in 9839256 — `ApexClassOptions` is now a discriminated union. `sobjecttype` is `?: never` on all non-Batchable templates and only `?: string` on `Batchable`. The generator likewise only injects `sobjectType` when `template === 'Batchable'`.

mshanemc:
@peternhale addressed in 35a8e57. Both findings fixed:

1. **Empty-string bypass** — generator now uses `sobjecttype?.trim() || 'SObject'` (and `validateOptions` skips validation for empty/whitespace), so `''` and `'   '` fall back to `SObject` instead of rendering `Database.Batchable<>`.

2. **Missing validation** — added `CreateUtil.checkSObjectType`, called from `validateOptions` when `template === 'Batchable'`. Note I did **not** reuse `checkInputs` as suggested: it rejects `__`, which would wrongly reject valid custom objects like `My_Object__c` / `ns__Foo__c`. The new validator enforces SObject API-name rules (start with letter, word chars, no trailing underscore) while allowing custom/namespaced objects. `'My Object'` and `'123Invalid'` now throw.

Added tests: custom `__c` object, empty-string default, invalid-name rejection.

peternhale (approved):

